PR #13, opened 2016-11-19
Add CritToken struct

### LINKED ISSUE #11: Critical section token
What do you think about having a struct like zinc's [NoInterrupts](https://github.com/hackndev/zinc/blob/master/src/hal/cortex_common/irq.rs#L23) struct and passing it into the closure for the `interrupt::free` function? The idea is to make a zero-sized struct that can't be created by users and is only ever created in a critical section. Then, if a function should only be called in a critical section it can take a reference to that struct which will guarantee that it will only be called in a critical section.

It would look something like this
```
mod interrupts {
    pub fn disable() {}
    pub fn enable() {}

    pub struct CritToken {
        _private: ()
    }

    pub fn free<F>(f: F)
        where F: FnOnce(&CritToken)
    {
        disable();

        f(&CritToken{ _private: () });

        enable();
    }
}

fn dangerous_func(tok: &interrupts::CritToken) {}

fn calling_dangerous_func() {
    interrupts::free(|tok| {
        dangerous_func(tok);
    });
}
```

awelkie:
This closes #11 

I just tested this locally. If you don't mind having a dev-dependency on compiletest-rs, I can add a test to demonstrate that you can't instantiate a `CritToken`, but I don't know how useful that would be...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- `interrupt::CritToken`, a struct that only exists in a critical section.
+
+### Changed
+
+- The closure that `interrupt::free` takes now takes a reference to an `interrupt::CritToken`.
+
 ## [v0.1.5]
 
 ### Added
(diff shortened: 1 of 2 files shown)


homunkulus:
:umbrella: The latest upstream changes (presumably 50779f2) made this pull request unmergeable. Please resolve the merge conflicts.

japaric:
done in #17

PR #22, opened 2017-04-30
Implement a complete high-level wrapper for SYST

whitequark:


japaric:
LGTM. I would certainly prefer something more svd2rust like since that API can result in less load / stores instructions when several operations need to be performed. But since this API can be added in a backward compatible fashion, I'd be fine with adding this API now and the svd2rust one later (with a new minor cortex-m version)

Thoughts @perlindgren? Would this API work for you SysTick-y stuff?

japaric:
closing in favor of #25

PR #43, opened 2017-06-12
Issue DSB; ISB after changing FPU status to ensure completion

whitequark:
This is necessary according to ARM, I suppose if some VFP op got scheduled right after an inlined FPU enable it would be trouble.

### DIFF
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -11,6 +11,7 @@ use core::ptr;
 use volatile_register::{RO, RW, WO};
 
 use interrupt::{CriticalSection, Nr};
+use asm;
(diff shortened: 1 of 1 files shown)


japaric:
> This is necessary according to ARM

Could you add a comment with a link to the documentation that states this?

Are these instructions also required when disabling the FPU? If so, shouldn't they be called *before* disabling the FPU? Wondering if the pipeline may execute the CPACR write while some floating point op is in progress or something like that.

whitequark:
Actually now that I look again, Cortex-M4 TRM does *not* say that DSB/ISB are required:
screenshot_20170617_105917

PR #69, opened 2017-12-22
revise peripheral API

### LINKED ISSUE #67: Update method signatures to better match scoped singletons
Now that scoped singletons are a thing, it makes sense to change some method signatures from taking `&self` to  taking `&mut self`. An example of this is the NVIC API. See comments by @japaric in https://github.com/japaric/cortex-m/pull/62.

I plan to take a look at this, but I won't be able to any time soon. If anyone wants to pick this up, go ahead!

japaric:
This PR changes the signature of many of the high level methods available on peripherals like
`NVIC::get_priority`. Additionally some instance methods have been turned into static methods. The
following guidelines have been used to apply the changes:

- If the method body contains a single, atomic read operation with *no* side effects (e.g. the read
  operation clears one of the bits of the register): the signature changed to make the method
  static, i.e. `&self` was removed from the signature.

- If the method involves writing to or a RMW operation on a register: the signature changed to take
  the singleton by `&mut self` reference.

- If the method involves only read operations where at least one of them modifies the value
  of a register: the signature changed to take the singleton by `&mut self` reference.

The rationale for this last guideline is that using `&self`, instead of `&mut self`, lets the user
(unintentionally) break abstractions in the presence of generators. Example below:

``` rust
let peripherals = Peripherals::take().unwrap();
let syst = &peripherals.SYST;

// tasks
let mut a = || {
    loop {
        // yielding "busy wait"
        while !a.has_wrapped() {
            yield;
        }

        // do stuff
    }
};

let mut b = || {
    // ..

    // *NOTE* the problem is in the line below: this `is_counter_enabled` method reads the CSR
    // register and that read operation clears the COUNTFLAG bit of the register (if set), which is
    // the bit the `has_wrapped` method checks for.
    if syst.is_counter_enabled() {
        // ..
    }

    // ..
};
```

One more guideline was considered but the required conditions didn't apply to any of the existing
methods:

- If the method involves only non side effectful, non necessarily atomic read operations: the
  signature of the method should remain as `&self`.

The rationale for this guideline is that a static method (no `self` argument) wouldn't be
appropriate because that can result in a torn read if the read operation can be preempted by some
context that modifies the register.

In any case, this last guideline doesn't seem to apply well to the peripherals structs exposed by
this crate because they *deref* to a `RegisterBlock` that allows mutation through a `&self`
reference. When these two properties (the guideline and `Deref<Target=RegisterBlock>`) are mixed
the user can potentially break abstractions using generators (as shown in the `syst` example).

cc @hannobraun
closes #67

### DIFF
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -7,10 +7,14 @@
 //! them at any given point in time) and the only way to get an instance of them is through the
 //! [`Peripherals::take`](struct.Peripherals.html#method.take) method.
 //!
-//! ``` ignore
+//! ``` no_run
+//! extern crate cortex_m;
+//!
+//! use cortex_m::peripheral::Peripherals;
+//!
 //! fn main() {
-//!     let peripherals = Peripherals::take();
-//!     peripherals.dwt.enable_cycle_counter();
+//!     let mut peripherals = Peripherals::take().unwrap();
+//!     peripherals.DWT.enable_cycle_counter();
 //! }
 //! ```
 //!
@@ -21,15 +25,19 @@
 //! API is provided as static methods on the peripheral types. One example is the
 //! [`DWT::cyccnt`](struct.DWT.html#method.cyccnt) method.
 //!
-//! ``` ignore
+//! ``` no_run
+//! extern crate cortex_m;
+//!
+//! use cortex_m::peripheral::{DWT, Peripherals};
+//!
 //! fn main() {
 //!     {
-//!         let peripherals = Peripherals::take().unwrap();
+//!         let mut peripherals = Peripherals::take().unwrap();
 //!         peripherals.DWT.enable_cycle_counter();
 //!     } // all the peripheral singletons are destroyed here
 //!
 //!     // but this method can be called without a DWT instance
-//!     let cyccnt = DWT::cyccnt();
+//!     let cyccnt = DWT::get_cycle_count();
 //! }
 //! ```
 //!
@@ -37,10 +45,14 @@
 //! available on all the peripheral types. This method is a useful building block for implementing
 //! higher level and safe abstractions.
 //!
-//! ``` ignore
+//! ``` no_run
+//! extern crate cortex_m;
+//!
+//! use cortex_m::peripheral::{DWT, Peripherals};
+//!
 //! fn main() {
 //!     {
-//!         let peripherals = Peripherals::take().unwrap();
+//!         let mut peripherals = Peripherals::take().unwrap();
 //!         peripherals.DWT.enable_cycle_counter();
 //!     } // all the peripheral singletons are destroyed here
 //!
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -1,5 +1,54 @@
 //! Core peripherals
 //!
+//! # API
+//!
+//! To use (most of) the peripheral API first you must get an *instance* of the peripheral. All the
+//! core peripherals are modeled as singletons (there can only ever be, at most, one instance of
+//! them at any given point in time) and the only way to get an instance of them is through the
+//! [`Peripherals::take`](struct.Peripherals.html#method.take) method.
+//!
+//! ``` ignore
+//! fn main() {
+//!     let peripherals = Peripherals::take();
+//!     peripherals.dwt.enable_cycle_counter();
+//! }
+//! ```
+//!
+//! This method can only be successfully called *once* -- this is why the method returns an
+//! `Option`. Subsequent calls to the method will result in a `None` value being returned.
+//!
+//! A part of the peripheral API doesn't require access to a peripheral instance. This part of the
+//! API is provided as static methods on the peripheral types. One example is the
+//! [`DWT::cyccnt`](struct.DWT.html#method.cyccnt) method.
+//!
+//! ``` ignore
+//! fn main() {
+//!     {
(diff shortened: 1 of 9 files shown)


homunkulus:
:umbrella: The latest upstream changes (presumably #71) made this pull request unmergeable. Please resolve the merge conflicts.

src/peripheral/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+//! core peripherals are modeled as singletons (there can only ever be, at most, one instance of
+//! them at any given point in time) and the only way to get an instance of them is through the
+//! [`Peripherals::take`](struct.Peripherals.html#method.take) method.
+//!
+//! ``` ignore
+//! fn main() {
+//!     let peripherals = Peripherals::take();
+//!     peripherals.dwt.enable_cycle_counter();
+//! }
+//! ```
```
hannobraun: This example is broken. Fixed version:
```
//! ``` no_run
//! use cortex_m::peripheral::Peripherals;
//!
//! fn main() {
//!     let mut peripherals = Peripherals::take().unwrap();
//!     peripherals.DWT.enable_cycle_counter();
//! }
//! ```
```

src/peripheral/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+//! fn main() {
+//!     {
+//!         let peripherals = Peripherals::take().unwrap();
+//!         peripherals.DWT.enable_cycle_counter();
+//!     } // all the peripheral singletons are destroyed here
+//!
+//!     // but this method can be called without a DWT instance
+//!     let cyccnt = DWT::cyccnt();
+//! }
+//! ```
```
hannobraun: This example is also broken. Fixed version:
```
//! ``` no_run
//! use cortex_m::peripheral::{
//!     DWT,
//!     Peripherals,
//! };
//!
//! fn main() {
//!     {
//!         let mut peripherals = Peripherals::take().unwrap();
//!         peripherals.DWT.enable_cycle_counter();
//!     } // all the peripheral singletons are destroyed here
//!
//!     // but this method can be called without a DWT instance
//!     let cyccnt = DWT::get_cycle_count();
//! }
//! ```
```

src/peripheral/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+//! fn main() {
+//!     {
+//!         let peripherals = Peripherals::take().unwrap();
+//!         peripherals.DWT.enable_cycle_counter();
+//!     } // all the peripheral singletons are destroyed here
+//!
+//!     // actually safe because this is an atomic read with no side effects
+//!     let cyccnt = unsafe { (*DWT::ptr()).cyccnt.read() };
+//! }
+//! ```
```
hannobraun: Also broken. Fixed version:
```
//! ``` no_run
//! use cortex_m::peripheral::{
//!     DWT,
//!     Peripherals,
//! };
//!
//! fn main() {
//!     {
//!         let mut peripherals = Peripherals::take().unwrap();
//!         peripherals.DWT.enable_cycle_counter();
//!     } // all the peripheral singletons are destroyed here
//!
//!     // actually safe because this is an atomic read with no side effects
//!     let cyccnt = unsafe { (*DWT::ptr()).cyccnt.read() };
//! }
//! ```
```

src/peripheral/nvic.rs
```diff
@@ -83,64 +84,69 @@ impl RegisterBlock {
...
 
-    /// Gets the "priority" of `interrupt`
+    /// Returns the NVIC priority of `interrupt`
     ///
-    /// NOTE NVIC encodes priority in the highest bits of a byte so values like
-    /// `1` and `2` have the same priority. Also for NVIC priorities, a lower
-    /// value (e.g. `16`) has higher priority than a larger value (e.g. `32`).
-    pub fn get_priority<I>(&self, interrupt: I) -> u8
+    /// *NOTE* NVIC encodes priority in the highest bits of a byte so values like `1` and `2` map
+    /// the same priority. Also for NVIC priorities, a lower value (e.g. `16`) has higher priority
```
hannobraun: Should be "map **to** the same priority".

japaric:
@hannobraun Thanks for the review!

@homunkulus r+

homunkulus:
:pushpin: Commit 5a19c39 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 5a19c39edc5a06f56c1ba61537d2fb61bf2ac818 with merge 08c13205a598e6e92870d499df95a4c18e5b9b30...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cortex-m/builds/327796853?utm_source=github_status&utm_medium=notification)

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit 47623ce has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 47623cebbbef76ccf0b5821a8b3213d4fa0a0c03 with merge 024527b41cdf3f5c9556b5eb94a1a4d2b33758ef...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cortex-m/builds/327800895?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 024527b41cdf3f5c9556b5eb94a1a4d2b33758ef to master...

PR #75, opened 2018-01-15
better document the SYST API

### LINKED ISSUE #59: feedback on the SYST API
"Some nits about the SYST implementation: fn clear () does not make a whole lot of sense as clearing actually disables the SysTick timer because the exception is triggered when the count changes from 1 -> 0 which is also what causes the reload. Similarly the manuals suggest you should load the counter with the value 1 (for which there's currently no safe API) to ensure that the timer will not start with some offset in the first turn and also the correct value for the reload is actually - 1."

from https://github.com/japaric/cortex-m-quickstart/issues/15#issuecomment-314676282

cc @whitequark

japaric:
closes #59

### DIFF
--- a/src/peripheral/syst.rs
+++ b/src/peripheral/syst.rs
@@ -39,8 +39,7 @@ const SYST_CALIB_NOREF: u32 = 1 << 31;
 impl SYST {
     /// Clears current value to 0
     ///
-    /// After calling `clear_current()`, the next call to `has_wrapped()`
-    /// will return `false`.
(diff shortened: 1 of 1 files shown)


japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit f0919f2 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit f0919f2b029d7a4c1d7cfc71fa252daba90eaf17 with merge bc31511494082273b90def2c3348f591ecb5b103...

PR #83, opened 2018-02-23
Change MenManage to MemManage

docbrown:


japaric:
superseded by #88 which renames the exception to `MemoryManagement`, which matches the CMSIS name.

PR #84, opened 2018-03-12
add asm::delay

japaric:
This PR adds an `asm::delay` function that can be used to block the program for the specified number
of instruction cycles (NB in absence of interrupts, see API docs). The implementation is done in
assembly so the execution time of that function is the same regardless of the used optimization
level.

Rationale: external devices (sensors) sometimes require delays in their initialization, or to
workaround silicon errata. Having to set up a hardware peripheral like the SysTick can be overkill
when the delay is to be done once.

Why not provide an implementation of `embedded_hal::{DelayUs, DelayMs}`? Going from instruction
cycles to human time requires knowing the clock configuration and in some cases the Flash access
latency. As all that information is device specific it's best left to device crates to implement.

Thoughts? Perhaps this is too error prone due to Flash access latency and interrupts? The interrupt
issue could be avoided by calling the asm! block within an interrupt::free call.

cc @therealprof @kunerd

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -24,6 +24,30 @@ pub fn bkpt() {
     }
 }
 
+/// Blocks the program for at least `n` instruction cycles
+///
+/// This is implemented in assembly so its execution time is the same regardless of the optimization
+/// level.
+///
+/// NOTE that the delay can take much longer if interrupts are serviced during its execution.
+#[inline(never)]
+pub fn delay(_n: u32) {
+    match () {
+        #[cfg(target_arch = "arm")]
+        () => unsafe {
+            asm!("1:
+                  subs $0, $$1
+                  bne.n 1b"
+                 :
+                 : "r"(_n / 3 + 1)
+                 :
+                 : "volatile");
+        },
+        #[cfg(not(target_arch = "arm"))]
+        () => unimplemented!(),
+    }
+}
+
 /// A no-operation. Useful to prevent delay loops from being optimized away.
 #[inline]
 pub fn nop() {
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -24,26 +24,37 @@ pub fn bkpt() {
     }
 }
 
-/// Blocks the program for at least `n` instruction cycles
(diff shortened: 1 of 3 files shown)


japaric:
bors try

src/asm.rs
```diff
@@ -15,6 +15,30 @@ pub fn bkpt() {
     }
 }
 
+/// Blocks the program for at least `n` instruction cycles
+///
+/// This is implemented in assembly so its execution time is the same regardless of the optimization
+/// level.
+///
+/// NOTE that the delay can take much longer if interrupts are serviced during its execution.
+#[inline(never)]
```
japaric: Without `#[inline(never)]` the compiler seems to misoptimize this function: it doesn't reload the `$0` register on every invocation but instead it starts decrementing it from its previous value (0).

pftbest: This is not misoptimized, it is intended behavior. You used the "r" constraint and promised to the compiler that you will not modify the value in this register. There is a way to fix it, use "r+" modifier to let compiler know that you will write this input register. Or you can copy the value in some other register, add it to the clobber list and modify it instead of input register.

therealprof:
That's a goodie. Far better than the loops I've been using every now and then. I agree with @pftbest that you'll need to change the register clobbers to make it not be inline, also I'd prefer something a bit more obvious than a delay calculated in times 3 cycles. Maybe throw in a NOP for good measure to make it stretch it a cycle to a more obvious 4?

I don't think interrupts are going to be a problem. Someone using this mechanism usually only cares about a **minimum** delay, not absolute precision.

kunerd:
I see why it doesn't make sense to implement it in embedded-hal, but I also have done some research and the requirement for small delays seems not to be so uncommon as I first thought. Would it make sense to have something like ['BogoMips'](https://en.wikipedia.org/wiki/BogoMips) and corresponding busy waiting traits in embedded-hal, like they are used in the Linux Kernel? I'm not an OS or experienced embedded guy, so I might be totally wrong with this.

japaric:
addressed the review comments

bors r+

PR #86, opened 2018-04-12
Add SEV assembly instruction

Nemo157:


japaric:
Thanks, @Nemo157!

bors r+

PR #88, opened 2018-04-25
make compilable on stable

japaric:
This PR makes this crate compilable on stable when the "inline-asm" and "singleton" Cargo features
are disabled (they are enabled by default to maintain backwards compatibility).

The main change has been replacing almost (\*) all inline `asm!` invocations with FFI calls into
external assembly files.

(\*) Stuff that has not been converted into external assembly file and thus is not available on
stable:

- Reading the (A)PSR register (I'm not sure if this will work with the extra function call overhead)

- Reading and writing the Link Register (LR)

- Reading and writing the Program Counter (PC)

I would appreciate if someone checked that all the stuff that's now using FFI calls has the same
semantics as the inline `asm!` version.


japaric:
> Stuff that has not been converted into external assembly file and thus is not available on
stable:

Given that this list is very short and the stuff in there is not commonly used I think it makes rust-lang-nursery/embedded-wg#63 much less urgent.

cc @jcsoo @thejpster @jamesmunns

asm/nop.s
```diff
@@ -0,0 +1,4 @@
+  .global __nop
+__nop:
+  nop
```
japaric: We could drop the `nop` instruction here. Even without it `asm::nop` won't be optimized away by the compiler.

src/interrupt.rs
```diff
@@ -1,16 +1,29 @@
...
         },
-        #[cfg(not(target_arch = "arm"))]
+
+        #[cfg(all(cortex_m, not(feature = "inline-asm")))]
+        () => unsafe {
+            extern "C" {
+                fn __cpsid();
+            }
+
+            // XXX do we need a explicit compiler barrier here?
```
japaric: My experiments indicate that FFI calls behave the same as `asm!("" ::: "memory" : "volatile")` (I suppose because the code that's being called could be doing an assembly call with those clobbers (?)) so I think `compiler_barrier` is not necessary here.

PR #91, opened 2018-05-14
Add STIR register to NVIC peripheral

thenewwazoo:
Quick and dirty PR to add software interrupt request capability

### DIFF
--- a/src/peripheral/nvic.rs
+++ b/src/peripheral/nvic.rs
@@ -2,7 +2,7 @@
 
 #[cfg(not(armv6m))]
 use volatile_register::RO;
-use volatile_register::RW;
+use volatile_register::{RW, WO};
 
 use interrupt::Nr;
 use peripheral::NVIC;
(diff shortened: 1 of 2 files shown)


src/peripheral/nvic.rs
```diff
@@ -65,9 +65,30 @@ pub struct RegisterBlock {
...
     pub ipr: [RW<u32>; 8],
+
+    #[cfg(not(armv6m))]
+    reserved5: [u32; 208],
+
+    #[cfg(armv6m)]
+    reserved5: [u32; 696],
+
+    /// Software Trigger Interrupt
+    pub stir: WO<u32>,
```
japaric: STIR is not present on ARMv6-M so this needs `#[cfg(not(armv6m))]`.

src/peripheral/nvic.rs
```diff
@@ -65,9 +65,30 @@ pub struct RegisterBlock {
...
 }
 
 impl NVIC {
+    /// Request an IRQ in software
+    pub fn req_irq<I>(&mut self, interrupt: I)
```
japaric: I think this method should be renamed to `request`. The other NVIC methods like `enabled(interrupt: I)` are read as "enable interrupt", so I think "request interrupt" would be more fitting.

I think this method should include a comment similar to what's included in the ARM documentation:

> Writing a value to the INTID field is the same as manually pending an interrupt by setting the corresponding interrupt bit in an Interrupt Set Pending Register.

japaric: > I think this method should include a comment similar to what's included in the ARM documentation:

in particular the comment should mention that this method is similar to `set_pending`.

japaric:
Thank you for this PR @thenewwazoo ! Unfortunately we haven't heard from you on this in a while, so I'm closing the PR to keep things tidy. Don't worry though, if you'll have time again in the future please **reopen** this PR, we'll be happy to review it again!

PR #94, opened 2018-06-17
Add method to trigger system reset

bergus:
Basically taken from CMSIS


little-arhat:
Linking https://github.com/japaric/cortex-m-rtfm/issues/77 and https://github.com/japaric/stm32f30x-hal/issues/24

japaric:
Thanks for the PR, @bergus. This LGTM but rustc is still affected by a LLVM "bug" (actually a C feature) where it misoptimizes infinite loops that have no side effects (see rust-lang/rust#28728) so the `loop {}` should be turned into `loop { NOP }` to avoid the bug.

PR #103, opened 2018-08-07
mark all assembly functions as .thumb_func

japaric:
this works around a [LLD bug] related to interworking.

[LLD bug]: https://bugs.llvm.org/show_bug.cgi?id=38435


adamgreig:
Ah thanks, I was misled by the old Homu docs I think!

bors r+

Issue #100, opened 2018-07-26
SCB missing VTOR on Cortex-M0+


gkelly:
The SCB definition uses `cfg(not(armv6m))` to choose whether or not to expose the `VTOR` field on non-armv6m platforms.  However, Cortex-M0+ has a `VTOR` field: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0662b/CIHFDJCA.html

Since Cortex-M0 doesn't have a `VTOR` field in its `SCB` block I'm not sure what the correct solution here is.  How do you feel about a feature that enables M0+ registers?  It's not as elegant as architecture detection through the target.

japaric:
@rust-embedded/cortex-m this needs a decision

> How do you feel about a feature that enables M0+ registers?

Cargo features are notoriously easy to get wrong because if the dependency enables it then there's no way to disable it.

Here, I would vote for the simplest solution of always exposing VTOR for the `thumbv6m-none-eabi` target and simply document that it's present on Cortex-M0+ but not on the Cortex-M0.

adamgreig:
+1 on always exposing VTOR.

therealprof:
+1

japaric:
Solution landed in #104

PR #104, opened 2018-08-08
Always expose the VTOR register (#100)

gkelly:
Cortex-M0+ platforms had the VTOR register hidden because of the way the
SCB chose to expose it.  This change exposes it on all platforms, with
the side effect that it will be visible on Cortex-M0 even thought the M0
SCB does not define it.


japaric:
Thanks @gkelly!

The team discussed this in #100 and there was consensus on implementing this solution.

bors r+

PR #106, opened 2018-08-14
Stir register and debugger check

thenewwazoo:
Adds support for requesting an interrupt via the STIR register and checking whether a debugger is attached.

### DIFF
--- a/src/peripheral/dcb.rs
+++ b/src/peripheral/dcb.rs
@@ -34,13 +34,16 @@ impl DCB {
         // unset bit 24 / TRCENA
         unsafe { self.demcr.modify(|w| w & !DCB_DEMCR_TRCENA); }
 
-    /// Is there a debugger attached? (see note)
+    /// Is there a debugger attached? (see notes)
     ///
-    /// N.B. This function is [reported not to
+    /// Note 1: This function is [reported not to
     /// work](http://web.archive.org/web/20180821191012/https://community.nxp.com/thread/424925#comment-782843)
     /// on Cortex-M0 devices. Per the ARM v6-M Architecture Reference Manual, "Access to the DHCSR
     /// from software running on the processor is IMPLEMENTATION DEFINED". Indeed, from the
     /// [Cortex-M0+ r0p1 Technical Reference Manual](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0484c/BABJHEIG.html), "Note Software cannot access the debug registers."
+    ///
+    /// Note 2: This function reads the DHCSR register, and therefore clears S_RESET_ST and
+    /// S_RETIRE_ST.
     pub fn is_debugger_attached() -> bool {
         unsafe { (*Self::ptr()).dhcsr.read() & 0x1 == 1 }
     }
--- a/src/peripheral/dcb.rs
+++ b/src/peripheral/dcb.rs
@@ -20,7 +20,7 @@ pub struct RegisterBlock {
 }
 
 impl DCB {
-    /// Enables TRACE. This is for example required by the 
+    /// Enables TRACE. This is for example required by the
     /// `peripheral::DWT` cycle counter to work properly.
     /// As by STM documentation, this flag is not reset on
     /// soft-reset, only on power reset.
@@ -28,10 +28,14 @@ impl DCB {
         // set bit 24 / TRCENA
         unsafe { self.demcr.modify(|w| w | DCB_DEMCR_TRCENA); }
     }
-    
+
     /// Disables TRACE. See `DCB::enable_trace()` for more details
     pub fn disable_trace(&mut self) {
         // unset bit 24 / TRCENA
         unsafe { self.demcr.modify(|w| w & !DCB_DEMCR_TRCENA); }
+
+    /// Is there a debugger attached?
+    pub fn is_debugger_attached(&self) -> bool {
+        self.dhcsr.read() & 0x1 == 1
     }
 }
--- a/src/peripheral/dcb.rs
+++ b/src/peripheral/dcb.rs
@@ -2,6 +2,7 @@
 
 use volatile_register::{RW, WO};
 
+use core::ptr;
 use peripheral::DCB;
 
 const DCB_DEMCR_TRCENA: u32 = 1 << 24;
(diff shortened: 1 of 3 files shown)


thenewwazoo:
(This is a re-submittal of #91, which I couldn't figure out how to re-open, sorry)

korken89:
bors try

korken89:
There were some errors in the build, for nightly it seems to be in `bare-metal` can you open an issue for this?

src/peripheral/nvic.rs
```diff
@@ -65,9 +65,36 @@ pub struct RegisterBlock {
...
+    pub stir: WO<u32>,
 }
 
 impl NVIC {
+    #[cfg(not(armv6m))]
+    /// Request an IRQ in software
+    ///
+    /// Writing a value to the INTID field is the same as manually pending an interrupt by setting
+    /// the corresponding interrupt bit in an Interrupt Set Pending Register. This is similar to
+    /// `set_pending`.
```
japaric: Could you include in the documentation why would one prefer this over `set_pending` and also mention that this API is not available on ARMv6-M.

thenewwazoo: I'll be honest... I don't know why to choose one vs the other. The reference manuals don't make the different clear. :) I'll add a note about not being available on armv6m

japaric:
@thenewwazoo did you get the `is_debugger_attached` method to work? i.e. is this ready for review? Actually, I'm going to let @korken89 finish the review. :-)

korken89:
I will give the PR a spin tonight :) If I find any issue I will report back!

thenewwazoo:
I was not able to get the `is_debugger_attached` method to work, but @adamgreig reported the same approach has historically worked for him on other devices than I've got on-hand. It's one of those things that _should_ work... In any case, I don't believe I'm missing any preconditions.

korken89:
I have it a quick test, and it's not working for me. I will dig deeper tomorrow and see what's missing.

adamgreig:
I've just tested this on a STM32F407 and it works as expected. In my case I flash an LED while debugger is attached and don't if not. I'll see if I can find a few other targets to test it on this week.

However I suggest replacing it with:

```rust
impl DCB {
    /// Is there a debugger attached?
    pub fn is_debugger_attached() -> bool {
        unsafe { (*Self::ptr()).dhcsr.read() & 0x1 == 1 }
    }
}
```

to make it accessible without an instance of `DCB`, just like `DWT::get_cycle_count()` and friends.

japaric:
> However I suggest replacing it with:

That'd be fine iff reading DHCSR does *not* clear any of its bits. (NB I haven't read this register documentation so I'm just asking for caution rather than saying this should not be done)

japaric:
If this is not working really perhaps if one of those optional things that vendors may or may not implement? If that's the case it should be noted in the docs.

adamgreig:
Checking the ARM, for DHCSR,

> Access to the DHCSR from software running on the processor is IMPLEMENTATION
DEFINED

And while this technique is [suggested by ARM](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka14536.html) (i.e., there's no other 'correct' way to do it AFAIK), it also [appears that](https://community.nxp.com/thread/424925#comment-782843) on Cortex-M0 the CPU can't read this register.

> That'd be fine iff reading DHCSR does not clear any of its bits.

Annoyingly S_RESET_ST and S_RETIRE_ST _are_ cleared by reading DHCSR, though I'm not sure if the CPU reading it will trigger the clear or only a debugger reading it. In any event S_RETIRE_ST is set to 1 whenever the processor completes an instruction, so just reading DHCSR would set the bit anyway. I wonder if it might be justifiable for this scenario. It's not like the exclusive access ensured by having a DCB object in any way synchronises access with the debugger anyway.

thenewwazoo:
Great find, @adamgreig!

It looks like I can gate this on `armv6m`, and build a more robust test for cortex-m0 targets in a separate crate. I also agree that making the `is_debugger_attached` method into a function would be better.

adamgreig:
Even on armv6m I'm not sure it's actually never going to work, just it seems like most Cortex-M0 don't permit it. I can't find anything in the ARMv6-M ARM to say it's prohibited, just the same caution that it's implementation defined as on the ARMv7-M ARM.

> build a more robust test for cortex-m0 targets

Can this be done? I can't see any good way to check for this on Cortex-M0.

korken89:
@thenewwazoo Can you add these comments and I will merge this?

korken89:
Thanks for the quick addition @thenewwazoo !
There seems to be a problem with one of the tests, could you give it a quick look?

thenewwazoo:
I looked into the test, and I've got a proximate cause.

```rust
    assert_eq!(address(&nvic.stir), 0xE000EF00);
```

is the line that fails. `address(&nvic.stir)` evaluates to `0xe000e930`, but according to [the TRM](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0337e/BHCFCFDB.html), `0xE000EF00` is the correct value.

thenewwazoo:
The test was executing for armv6m targets too (where the register doesn't exist), so I added a `cfg` attribute to skip it. The current failure is when building for x86 targets, where it is similarly nonsensical.

I figure I can:
* also deny the test for x86 targets
* refactor the tests to only permit armv7m, armv7em, ...?
* adjust the reserved5 register values in `nvic.rs` so all platforms line up and the test "succeeds"
* leave it as-is

korken89:
Thanks for checking and finding the cause @thenewwazoo ! Checking the tests and the CI scripts I am not sure why the x86 target is run at all. @rust-embedded/cortex-m am I missing something obvious? Else I will opt for removing the `cargo test` and only keep the `cargo check` for the x86 target.

src/peripheral/dcb.rs
```diff
@@ -20,18 +20,32 @@ pub struct RegisterBlock {
...
+
+    /// Is there a debugger attached? (see notes)
+    ///
+    /// Note 1: This function is [reported not to
+    /// work](http://web.archive.org/web/20180821191012/https://community.nxp.com/thread/424925#comment-782843)
+    /// on Cortex-M0 devices. Per the ARM v6-M Architecture Reference Manual, "Access to the DHCSR
+    /// from software running on the processor is IMPLEMENTATION DEFINED". Indeed, from the
+    /// [Cortex-M0+ r0p1 Technical Reference Manual](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0484c/BABJHEIG.html), "Note Software cannot access the debug registers."
+    ///
+    /// Note 2: This function reads the DHCSR register, and therefore clears S_RESET_ST and
```
japaric: I'm still not happy about this being a static method that modifies the contents of the register. The documentation says:

> Writes to this register in any size other than word are Unpredictable. It is acceptable to read in any size, and you can use it to avoid or intentionally change a sticky bit.

Could you turn this into a byte read to avoid clearing the S_RESET_ST and S_RETIRE_ST bits? Then it'd be fine to keep this as a static method.

thenewwazoo: I've tried a couple of ways to do this, but am not sure which you'd prefer.

Byte-wise access:
```rust
pub struct RegisterBlock {
    /// Debug Halting Control and Status
    pub dhcsr: [RW<u8>;4],
...
```

breaks writing to the register, which requires that the high halfword always be `0xA05F`. 

Using an untagged union:
```rust
pub union DHCSR {
    /// 32-bit word access
    w: WO<u32>,
    /// Individual byte access
    b: [RW<u8>; 4],
}

/// Register block
#[repr(C)]
pub struct RegisterBlock {
    /// Debug Halting Control and Status
    pub dhcsr: DHCSR,
...
```
looks like a good solution, but untagged unions of non-Copy types is unstable.

I could try to do some unsafe trickery by doing pointer math but that seems like a bad idea. What do you think?

As an aside, I'm sorry this tiny little change has eaten up so much energy. It seemed so simple at first. :(

thenewwazoo: I was reminded of this today - any suggestions? I'm happy to go with whatever's best, I just don't know what it is. :)

japaric: @thenewwazoo sorry for dropping the ball here

> breaks writing to the register, which requires that the high halfword always be 0xA05F.

this would be a breaking change; also it breaks writes as you say

> Using an untagged union:

this is unstable so it's a no-go

> I could try to do some unsafe trickery by doing pointer math but that seems like a bad idea. 

using pointers seems like the only stable, backward-compatible solution. You can write something like `(*(&dhcsr as *const [RW<u8>; 4]))[index].read())` to avoid the pointer math. Though pointer math is probably easier to read `ptr::read_volatile((&dhcsr as *const u8).add(index))`.

thenewwazoo:
@Disasm @adamgreig this should be GTG (perhaps pending @japaric confirming his requested changes have been satisfied)

thenewwazoo:
Tests should succeed now, having fixed the reserved offsets

adamgreig:
thanks!

bors r+

PR #112, opened 2018-09-02
Add PendSV exception set and clear to SCB

ekohandel:
This change adds the ability to easily set, clear, and inquire the status of the PendSV exception via the SCB peripheral on all Cortex-M platforms.

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -615,3 +615,29 @@ impl SCB {
         }
     }
 }
+
+const SCB_ICSR_PENDSVSET: u32 = 1 << 28;
+const SCB_ICSR_PENDSVCLR: u32 = 1 << 27;
+
+impl SCB {
+    /// Set the PENDSVSET bit in the ICSR register which will pend the PendSV interrupt
+    pub fn set_pendsv() {
+        unsafe {
+            (*Self::ptr()).icsr.write(SCB_ICSR_PENDSVSET);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for the PR!

I think all three methods could be made into functions, in other words, without taking `self` but instead using `unsafe` to directly access ICSR. That allows you to call them without requiring possession of `SCB`. Since all bits in this register are either RO, WO, or RW with set-to-act semantics, I don't think you can get a data race with read-modify-write: we can always either read the value, or write a value with a single 1 bit set. In other words you could replace all your current `modify` calls with simple `write` of just the relevant bit. What do you think?

ekohandel:
Thank you for your feedback! That makes sense to me. I am very new to rust and followed some other patterns in the same file. As you mentioned the SET/CLR bit pattern ensures that there are no race conditions by design.

I have changed my PR to what you had suggested. I did enclose the unsafe operations within an `unsafe` block inside the functions to keep the functions themselves safe. Not sure if this would have any negative impacts or lead to any misleading information from an API perspective.

adamgreig:
Sorry, a Travis check timed out and I hadn't noticed. This looks good to me!

> I did enclose the unsafe operations within an unsafe block inside the functions to keep the functions themselves safe. Not sure if this would have any negative impacts or lead to any misleading information from an API perspective.

That's what I intended. Since writing 0 to the other fields has no effect, and you're not doing a read-modify-write at all, it should always be safe to call these functions (in that you won't trigger any UB).

bors r+

PR #115, opened 2018-09-06
make `iprintln!` not depend on `iprint!`

japaric:
the preferred way to import macros in Rust 2018 is via `use`. If you import
`iprintln` and try to use you'll get an error if the `iprint` macro has not been
imported as well.

This commit makes `iprintln` work w/o having to import `iprint` as well.

r? @rust-embedded/cortex-m (anyone)


japaric:
bors retry

network error: https://travis-ci.org/rust-embedded/cortex-m/jobs/425416906

PR #116, opened 2018-09-24
Add PendSV and SysTick flags

qwerty19106:
CMSIS core headers contains SCB_ICSR_PENDSVSET_*** and SCB_ICSR_PENDSTSET_*** definitions, which I need in my projects. In CMSIS it is used to check, set and clear this flags (rtx_core_cm.h and os_systick.c).

I suggest adding it to scb.rs. I put initial commit, but I need help to add compiler barriers where its are needed.

For details, see CMSIS:
[CMSIS_5/CMSIS/Core/Include/core_cm3.h](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/Core/Include/core_cm3.h)  (or other core_cm{X}.h)
[CMSIS_5/CMSIS/RTOS2/RTX/Source/rtx_core_cm.h)](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/RTOS2/RTX/Source/rtx_core_cm.h)
[CMSIS_5/CMSIS/RTOS2/Source/os_systick.c)](https://github.com/ARM-software/CMSIS_5/blob/develop/CMSIS/RTOS2/Source/os_systick.c)

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -297,6 +297,12 @@ impl VectActive {
 mod scb_consts {
     pub const SCB_CCR_IC_MASK: u32 = (1 << 17);
     pub const SCB_CCR_DC_MASK: u32 = (1 << 16);
+
+    pub const SCB_ICSR_PENDSVSET_MASK: u32 = 1 << 28;
+    pub const SCB_ICSR_PENDSVCLR_MASK: u32 = 1 << 27;
+
+    pub const SCB_ICSR_PENDSTSET_MASK: u32 = 1 << 26;
+    pub const SCB_ICSR_PENDSTCLR_MASK: u32 = 1 << 25;
 }
 
 #[cfg(not(armv6m))]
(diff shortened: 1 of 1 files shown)


qwerty19106:
FIX: I have not seen that PENDSV is already added. I review PENDST code to be like PENDSV.

src/peripheral/scb.rs
```diff
@@ -640,4 +643,29 @@ impl SCB {
...
+    pub fn is_systick_pending() -> bool {
+        unsafe {
+            (*Self::ptr()).icsr.read() & SCB_ICSR_PENDSTSET == SCB_ICSR_PENDSTSET
+        }
+    }
+
+    
+    /// Set the PENDSTCLR bit in the ICSR register which will clear a pending SysTick interrupt
+    #[inline]
+    pub fn clear_systick_pending(&mut self) {
```
adamgreig: Does this operation require exclusive access to SCB (`&mut self`)? With `clear_pendsv()` it wasn't required because setting just the `PENDSTCLR` bit doesn't affect any other bits in the register, so you can't have data races.

src/peripheral/scb.rs
```diff
@@ -640,4 +643,29 @@ impl SCB {
             (*Self::ptr()).icsr.write(SCB_ICSR_PENDSVCLR);
         }
     }
+
+    /// Set the PENDSTCLR bit in the ICSR register which will clear a pending SysTick interrupt
+    #[inline]
+    pub fn set_systick(&mut self) {
```
adamgreig: I'm not sure about calling this `set_systick` when just above we have `set_pendsv`. Really this one should be `set_pendst` or we need to rename both (causing a major version change). Same applies to `is_systick_pending` and `clear_systick_pending`. I'd prefer just using `pendst` here.

src/peripheral/scb.rs
qwerty19106: Yes, exclusive access to SCB is not required. I think that `clear_systick_pending` needs to change
 analogiously.

src/peripheral/scb.rs
qwerty19106: Yes, `set_pendst` looks better.

qwerty19106:
I added requested changes, needs to review.

src/peripheral/scb.rs
```diff
@@ -640,4 +643,29 @@ impl SCB {
...
+
+    /// Check if PENDSTSET bit in the ICSR register is set meaning SysTick interrupt is pending
+    #[inline]
+    pub fn is_pendst_pending() -> bool {
+        unsafe {
+            (*Self::ptr()).icsr.read() & SCB_ICSR_PENDSTSET == SCB_ICSR_PENDSTSET
+        }
+    }
+
+
```
therealprof: Minor nit: One blank line too many.

qwerty19106:
Do I need to run rustfmt before create commit?
I try run rustfmt and got many changes on other lines.

therealprof:
Not a hard requirement but IMHO a good thing to do.

qwerty19106:
Are there any other changes i have to make?

adamgreig:
@therealprof are you happy with the result of rustfmt here, especially around L610? If anything I think it looks a bit messier now. I'm happy with everything else in the PR.

therealprof:
@adamgreig I agree it looks messier now but I'm not in charge of the formatting rules. I still think adhering to the standard formatting is a good idea even if the result is not what I personally prefer.

adamgreig:
Okay, let's merge this and we can consider tidying up more of the code for rustfmt another time.

adamgreig:
Thanks @qwerty19106 !

PR #117, opened 2018-10-24
add high level API to set priority of system handlers

japaric:
needed for cortex-m-rtfm v0.4.x (it makes my life easier / it makes it easier to
support ARMv6-M)

r? @rust-embedded/cortex-m (anyone)

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -669,3 +669,116 @@ impl SCB {
         }
     }
 }
+
+/// System handlers, exceptions with configurable priority
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum SystemHandler {
+    // NonMaskableInt, // priority is fixed
+    // HardFault, // priority is fixed
+    /// Memory management interrupt (not present on Cortex-M0 variants)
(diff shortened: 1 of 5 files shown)


src/peripheral/scb.rs
```diff
@@ -669,3 +669,116 @@ impl SCB {
...
+    UsageFault,
+
+    /// Secure fault interrupt (only on ARMv8-M)
+    #[cfg(any(armv8m, target_arch = "x86_64"))]
+    SecureFault,
+
+    /// SV call interrupt
+    SVCall,
+
+    // #[cfg(not(armv6m))]
```
therealprof: I'm not so hot about having this commented out bit in here. Can we figure out whether it has configurable priority and either take it in or leave it out?

adamgreig: It's got `PRI_12` in `SHPR3` too:
image

japaric: @adamgreig Thanks. PRI_12 was omitted in the Cortex-M4 documentation I read. (Not the first time ARM omits information about debugging / profiling capabilities in some revisions of their docs)

japaric:
Addressed review comments

therealprof:
bors try

PR #124, opened 2018-11-20
Add api to set SLEEPONEXIT bit in SCR register

simonvandel:


korken89:
Thanks for the addition, I was almost 100% sure this was already in there :)

bors r+

PR #126, opened 2018-12-09
Upgrade to aligned 0.3

rnestler:


adamgreig:
Sorry, just occurred to me that this will require a minor version bump in cortex-m since the function type changes. It remains approved but we should hold off on merging until we're ready to do a minor release.

rnestler:
> Sorry, just occurred to me that this will require a minor version bump in cortex-m since the function type changes. It remains approved but we should hold off on merging until we're ready to do a minor release.

Yes it would be a breaking change. Should I have marked it in some way as that?

adamgreig:
We don't have such a convention (yet!), that was just on me not thinking it through. #127 is also going to be a breaking change so we can do both at the next release.

adamgreig:
Merging in preparation of releasing 0.6.0.

bors r+

PR #127, opened 2018-12-15
Cortex M0(+) DWT fixes

korken89:
The current DWT setup has a lot of registers that are not available in Cortex-M0(+), fixes are added here.

### DIFF
--- a/src/peripheral/dwt.rs
+++ b/src/peripheral/dwt.rs
@@ -1,6 +1,8 @@
 //! Data Watchpoint and Trace unit
 
-use volatile_register::{RO, RW, WO};
+#[cfg(not(armv6m))]
+use volatile_register::WO;
+use volatile_register::{RO, RW};
 
 use peripheral::DWT;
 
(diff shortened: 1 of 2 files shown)


korken89:
Detail for DWT can be found here: http://infocenter.arm.com/help/topic/com.arm.doc.ddi0432c/DDI0432C_cortex_m0_r0p0_trm.pdf

korken89:
There, I think this is ready. Comment/suggestions @rust-embedded/cortex-m ?

korken89:
Indeed, but it is a bug now so a break is inevitable.
I would opt for fixing this ASAP as people might think that the DWT cyccnt exists now.

therealprof:
I tend to agree with @korken89 here. This is preventing the use of undefined behaviour (most likely even a crash) so it has precedence over "a breaking change" and thus should go in ASAP.

adamgreig:
Merging in preparation of releasing 0.6.0.

bors r+

PR #132, opened 2019-02-06
add newline

japaric:
testing highfive


japaric:
cc @rust-embedded/all we have highfive support!

let's discuss which repositories to enable it on in the internal team / GH chat or a meeting.

PR #133, opened 2019-02-07
add thumbv8m.base to ci

evq:
baseline support was added in https://github.com/rust-embedded/cortex-m/pull/123 but it wasn't added to the travis config


evq:
oops! looks like this requires a PR to rust first to add thumbv8 baseline to the list of targets which get their dist components built

PR #140, opened 2019-04-24
Cargo.toml: Add `links = "cortex-m"`

### LINKED ISSUE #137: Consider using Cargo's `links` feature to ensure use only once
Currently cortex-m uses an [unmangled variable](https://github.com/rust-embedded/cortex-m/blob/be6c39c725d2ae0cf50e36e07a27fd9a882cb71d/src/peripheral/mod.rs#L150-L153) to prevent multiple crates being linked in that redefine CORE_PERIPHERALS.

Cargo now has a feature called `links` that [addresses this](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key).

jonas-schievink:
This prevents linking multiple versions of `cortex-m` together, which would be
unsound. Currently it uses a `#[no_mangle]` static for this, which isn't always
reliable.

Fixes #137

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ name = "cortex-m"
 readme = "README.md"
 repository = "https://github.com/japaric/cortex-m"
 version = "0.6.0"
+links = "cortex-m"  # prevent multiple versions of this crate to be linked together
 
 [dependencies]
 aligned = "0.3.1"


jonas-schievink:
Since this will only work for other `cortex-m` versions also using `links`, I kept the `#[no_mangle]` solution in place for now (it's better than nothing)

jonas-schievink:
This might be https://github.com/rust-lang/rust/issues/60050?

jonas-schievink:
@therealprof can you give this another go? The CI failure should've gone away.

PR #142, opened 2019-04-28
New 0.5.9 which re-exports the CORE_PERIPHERALS static from 0.6.0

adamgreig:
I'm proposing this PR (or that we do something similar) to help fix the current ecosystem breakage with the new cortex-m 0.6.0 release.

In this PR, we only share the `CORE_PERIPHERALS` static bool, which means whether or not peripherals are taken is shared but the actual types are still different (i.e. this is not quite the semver trick).

This simple option will not allow sharing types between crates using different versions of cortex-m, but does allow them to co-exist where that isn't required, which I think is the majority of current breakages. 

Specifically, this allows using cortex_m_semihosting and/or panic_semihosting with application code and PACs using any combination of cortex-m 0.5 and 0.6, along with similar use cases.

The alternative would be to more extensively use semver trick to re-export more types, but I'd rather avoid this unless we think it's actually necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,12 +8,13 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m"
 readme = "README.md"
 repository = "https://github.com/japaric/cortex-m"
-version = "0.5.8"
+version = "0.5.9"
 
 [dependencies]
 aligned = "0.2.0"
 bare-metal = "0.2.0"
 volatile-register = "0.2.0"
+cortex_m_0_6 = { package = "cortex-m", version = "0.6.0" }
 
 [features]
 cm7-r0p1 = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ keywords = ["arm", "cortex-m", "register", "peripheral"]
 license = "MIT OR Apache-2.0"
 name = "cortex-m"
 readme = "README.md"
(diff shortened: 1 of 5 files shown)


therealprof:
bors try

korken89:
Hmm, is it maybe better to keep the linking part in a separate crate that both 0.5 and 0.6 will use? Else we will have the same issue when 0.7 comes. I'm at the oxidize impl, @adamgreig we can discuss online here.

adamgreig:
Updated to also do the semver trick to re-export all the types we can.

adamgreig:
bors try

Cargo.toml
```diff
@@ -8,12 +8,13 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m"
 readme = "README.md"
 repository = "https://github.com/japaric/cortex-m"
```
therealprof: Care to change this while you are at it?

adamgreig:
Force pushed with updated README and fixing build error on thumbv6.

adamgreig:
bors try

Cargo.toml
adamgreig: Done

japaric:
:shipit:

adamgreig:
Published :tada:

therealprof:
And it works, too! 🕺

PR #145, opened 2019-05-07
scb: add static version of `set_sleepdeep` and `clear_sleepdeep` as `set_sleep_deep` and `clear_sleep_deep` respectively.

azasypkin:
Makes a change similar to the one made here https://github.com/rust-embedded/cortex-m/pull/138, but for `set_sleepdeep`/`clear_sleepdeep` method pair.

Didn't find any better alternative names as `set_sleep_deep` and `clear_sleep_deep` respectively, happy to hear other suggestions.

Fixes https://github.com/japaric/cortex-m-rtfm/issues/165.

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -582,16 +582,32 @@ const SCB_SCR_SLEEPDEEP: u32 = 0x1 << 2;
 
 impl SCB {
     /// Set the SLEEPDEEP bit in the SCR register
+    #[deprecated(since = "0.6.1", note = "Use `SCB::set_sleep_deep`")]
     pub fn set_sleepdeep(&mut self) {
         unsafe {
             self.scr.modify(|scr| scr | SCB_SCR_SLEEPDEEP);
         }
     }
 
+    /// Set the SLEEPDEEP bit in the SCR register
+    pub fn set_sleep_deep() {
+        unsafe {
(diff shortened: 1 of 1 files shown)


adamgreig:
Hi, and thanks for the PR!

I can see why you'd want a static `SLEEPDEEP` method, but I think there's a potential problem with this implementation. Unlike `sys_reset()`, calling `modify()` on SCR might race against other callers, causing bits to be lost. The new static methods don't have any synchronisation, so for example if you called `set_sleep_deep()`, it read SCR, then an interrupt fired which calls `modify()` to update some other bit in SCR, the interrupt writes its new value, execution returns to your code, which writes its version of SCR with `SLEEPDEEP` set, erasing the change the interrupt made.

For this sort of register there's just no good way around this problem. You would either need to enter a critical section (disable all other interrupts), or mark this method `unsafe` and let callers provide their own guarantees. I don't think we have any such methods in cortex-m at the moment, though in principle they could exist...

azasypkin:
Thanks for reply @adamgreig! Yeah, my bad, what you're saying makes total sense.

The main reason behind this PR was to somehow access SCB from an interrupt when using `cortex-m-rtfm` since currently it's only available in `init` function. It's definitely not a problem of `cortex-m` though, so I guess I'll just postpone migration to `cortex-m-rtfm` until we have a "safe" solution.

PR #147, opened 2019-06-13
Remove debug_assert from Peripherals::steal

mvirkkunen:
This is the same change as https://github.com/rust-embedded/svd2rust/pull/238, except for the one bit that isn't generated by svd2rust. There was a decent amount of discussion about this over at that issue, and to me it makes sense to change this here as well. It's `unsafe`, so let the user decide if they want to use it.


thejpster:
Makes sense to me.

bors r+

PR #159, opened 2019-07-24
Define MSRV to be 1.31.0 and unconditionally use const-fn

### LINKED ISSUE #153: Make const-fn default
At the moment the cortex-m still hides `const` versions of some functions (especially `Mutex::new()`) behind the `const-fn` feature, but these have worked on stable Rust for some time now.

It looks like this crate does not yet follow the WG [MSRV policy](https://github.com/rust-embedded/wg/blob/master/ops/msrv.md) so I'm not really sure what the MSRV even is, but perhaps we could allow it to be 1.31 and therefore either remove the `const-fn` feature or make it a default.

cc https://github.com/rust-embedded/book/pull/196

adamgreig:
As per https://github.com/rust-embedded/wg/blob/master/ops/msrv.md we should have a defined MSRV for this crate. This PR proposes setting it to 1.31 to allow use of `const-fn` from bare-metal by default, and updates `Cargo.toml` accordingly. It also sets the edition to 2018, permitted by an MSRV of 1.31.

Since this PR _introduces_ an MSRV I propose it not requiring a semver bump, and we instead would look to release this change as `0.6.1`.

Closes #153.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,10 +16,9 @@ links = "cortex-m"  # prevent multiple versions of this crate to be linked toget
 
 [dependencies]
 aligned = "0.3.1"
-bare-metal = "0.2.0"
+bare-metal = { version = "0.2.0", features = ["const-fn"] }
 volatile-register = "0.2.0"
 
 [features]
 cm7-r0p1 = []
-const-fn = ["bare-metal/const-fn"]
 inline-asm = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 repository = "https://github.com/japaric/cortex-m"
 version = "0.6.0"
 links = "cortex-m"  # prevent multiple versions of this crate to be linked together
+edition = "2018"
 
 [dependencies]
 aligned = "0.3.1"
@@ -20,5 +21,6 @@ bare-metal = { version = "0.2.0", features = ["const-fn"] }
 volatile-register = "0.2.0"
 
 [features]
+const-fn = []
 cm7-r0p1 = []
 inline-asm = []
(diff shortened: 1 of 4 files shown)


therealprof:
Weird. https://github.com/rust-embedded/wg/blob/master/ops/msrv.md claims `const-fn` requires Rust 1.34, but https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#const-fn disagrees. ;)

Cargo.toml
```diff
@@ -16,10 +16,9 @@ links = "cortex-m"  # prevent multiple versions of this crate to be linked toget
 
 [dependencies]
 aligned = "0.3.1"
-bare-metal = "0.2.0"
+bare-metal = { version = "0.2.0", features = ["const-fn"] }
 volatile-register = "0.2.0"
 
 [features]
 cm7-r0p1 = []
-const-fn = ["bare-metal/const-fn"]
```
therealprof: I think this should rather be:
`const-fn = []`
otherwise I'd be a breaking change for people using the crate since you proposed not to bump the minor version.

adamgreig: :+1:, updated

adamgreig:
> Weird. https://github.com/rust-embedded/wg/blob/master/ops/msrv.md claims `const-fn` requires Rust 1.34, but https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#const-fn disagrees. ;)

I think this is just meant to be an example of a feature using a higher MSRV rather than specifically written to indicate const-fn needs 1.34.

adamgreig:
@therealprof I've updated this PR to remove the breaking edition-2018 stuff, which I'll do in another PR. This now passes CI.

PR #160, opened 2019-07-29
Update for 2018 edition

adamgreig:


korken89:
Thanks for the PR!
It seems from CI that some `crate::` is missing, can you have a look?

adamgreig:
Hah, I knew CI would catch something. Cargo fix missed it too. Anyway, updated, let's try again...

korken89:
Thanks, LGTM!

PR #161, opened 2019-08-11
Add {M,P}SPLIM access routines found on ARMv8-M mainline extension

aurabindo:
I'd like to add better v8-M support. I'm not sure if this is PR has an optimal way of handling arch extension. As per ARMv8 Architecture reference manual, the baseline variant is not compatible with ARMv7-M. Only the main extension brings backward compatibility with v7-M.

This PR adds support for accessing MSPLIM and PSPLIM registers, and missing binaries for `thumbv8m.main-none-eabihf` architecture.


thejpster:
Thank you for this - good spot. I look forward to adding msplim support to cortex-m-rt.

Issue #58, opened 2017-07-11
Privileged mode


kjetilkjeka:
There is currently no support for working in privileged mode.

There should at least be
- a way to call the svc instruction
    - This can be implemented by ```cortex_m::asm::svc(num: u32)```
- a way to write thread mode privilege level
    - This can be implemented by ```cortex_m::register::control::write(Control)``` and perhaps also something in the lines of ```cortex_m::register::control::modify(f: F) where f: FnOnce(Control) -> Control```

The cortex_m_rt crate seem to already implement the possibility for overriding the SVC handler. So with these changed it should be possible to: 
- Execute the work in need of privileged mode from the svc handler
- Set the thread mode from the SVC handler and then execute the work after returning from the handler.

kanishkarj:
What is the status of this? Currently, I am having to write assembly code and create pendSV interrupt manually. I figured that out from [here](https://github.com/n-k/cortexm-threads/blob/master/src/lib.rs#L258). Is there any better way ?

korken89:
It is this you are referring to? https://docs.rs/cortex-m/0.6.0/cortex_m/peripheral/struct.SCB.html#method.set_pendsv
This will pend the SVCall exception.

korken89:
The true SVC instruction is a bit problematic, as this requires `asm!("svc 123" :::: "volatile")`, which is not allowed on stable and the `123` is not allowed to be a register which thwarts the current workaround of the `asm.s` file.

However, if you run the nightly compiler it should be no issue, you can use the functions here for reference https://github.com/rust-embedded/cortex-m/blob/master/src/asm.rs

Or if you want stable, create your own `my_asm.s` is which you could do something like this:
```
  .section .text.__svc_123
  .global __svc_123
  .thumb_func
__svc_123:
  svc 123
  bx lr
```

kanishkarj:
Oh yeah, I did come across those nightly compiler features. I guess I have to work with stable, so the second option I guess. Thank you :)

kanishkarj:
Hey, also is there any way to Run code in privileged mode. Also is there anyway to make your code use the Main stack pointer instead of the process stack pointer.

jonas-schievink:
@kanishkarj Privileged mode and Main Stack Pointer (MSP) are used by default after reset of Cortex-M processors. The issue title and description are incorrect, I'm assuming it's talking about entering **un**privileged mode (and using the Process Stack Pointer instead of MSP).

[...]

PR #164, opened 2019-09-06
Allow writing to the CONTROL register

jonas-schievink:
This allows entering unprivileged mode.

Part of https://github.com/rust-embedded/cortex-m/issues/58


korken89:
Hi, thanks for the PR!
CI is reporting an error with `__control_w`, can you have a look?

jonas-schievink:
Fixed. I should properly test this, though...

jonas-schievink:
Okay, did some manual testing and it seems to work fine

asm.s
```diff
@@ -5,13 +5,21 @@ __bkpt:
...
-__control:
+__control_r:
   mrs r0, CONTROL
   bx lr
 
+  .section .text.__control_w
+  .global __control_w
+  .thumb_func
+__control_w:
+  msr CONTROL, r0
```
adamgreig: Do you think we should insert `ISB` here to ensure the write has taken effect? The architecture reference manual suggests always issuing `ISB` after a write to `CONTROL`.

src/register/control.rs
```diff
@@ -134,3 +180,30 @@ pub fn read() -> Control {
         () => unimplemented!(),
     }
 }
+
+/// Writes to the CPU register.
+#[inline]
+pub unsafe fn write(_control: Control) {
```
adamgreig: The use of `_control` as a name is a bit unconventional, because usually leading underscores indicate unused parameters. It's fine to call it `control` here and then shadow it below with a `let control = control.bits();`.

asm.s
adamgreig: (and if so, we should also put an ISB in the inline asm in `fn write()`).

jonas-schievink: Hmm, the Cortex-M0+ and Cortex-M4 User Guides only mention to use ISB after changing the SPSEL bit, not any of the other bits, so maybe this shouldn't be done unconditionally.

src/register/control.rs
jonas-schievink: Other register `write` functions do the same, since the value isn't used when the `#[cfg(not(cortex_m))]` path is active

kanishkarj:
Hey! i need the `set_npriv` badly for proceeding further with a project I am working on. Which version of `cortex-m` crate will support it?

jonas-schievink:
@kanishkarj Since this is not a breaking change it could be in 0.6.2

PR #171, opened 2019-10-09
Add `#[inline]` to lots of trivial functions.

m-ou-se:
Now the only public non-inline functions left are:

 - `write_all`
 - `write_aligned`
 - All (derived) `Debug` implementations

(Checked using Clippy's [`missing_inline_in_public_items`][1] lint.)

[1]: https://rust-lang.github.io/rust-clippy/master/#missing_inline_in_public_items

### DIFF
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -57,6 +57,7 @@ pub unsafe fn enable() {
 /// Execute closure `f` in an interrupt-free context.
 ///
 /// This as also known as a "critical section".
+#[inline]
 pub fn free<F, R>(f: F) -> R
 where
     F: FnOnce(&CriticalSection) -> R,
--- a/src/itm.rs
+++ b/src/itm.rs
@@ -21,6 +21,7 @@ unsafe fn write_words(stim: &mut Stim, bytes: &[u32]) {
 struct Port<'p>(&'p mut Stim);
 
 impl<'p> fmt::Write for Port<'p> {
+    #[inline]
     fn write_str(&mut self, s: &str) -> fmt::Result {
         write_all(self.0, s.as_bytes());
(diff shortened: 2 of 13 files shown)


m-ou-se:
`#![deny(clippy::missing_inline_in_public_items)]` can be added to prevent non-inline functions from accedentally being added.

#172 fixes (and/or silences) all other Clippy warnings, so Clippy can be enabled in CI.

korken89:
Overall this looks good! One question, have you done any test on the difference in size/perf. of optimized builds?
I am interested in what impact this has, even though it probably is minimal with the compiler doing a good job on inlining.

m-ou-se:
I have no numbers and statistics for you, but I did look at the disassembly of a few projects I'm working on. Before this change, lots of trivial functions ended up as calls to functions that do nothing other than return a constant or write or read a single register.

See also:

- https://github.com/rust-embedded/svd2rust/issues/287
- https://twitter.com/m_ou_se/status/1181877498474307584

m-ou-se:
Now with some numbers:

For some random stm32 project I'm working at:
```
$ arm-none-eabi-size before after 
   text	   data	    bss	    dec	    hex	filename
  11084	    256	    456	  11796	   2e14	before
  10888	    256	    456	  11600	   2d50	after
```

therealprof:
> Overall this looks good! One question, have you done any test on the difference in size/perf. of optimized builds?

Also note that upstream Rust is doing the exact same thing for trivial functions. cf. https://github.com/rust-lang/rust/pull/64996

Especially the `#[inline(always)]` on trivial functions help a lot on debug builds.

korken89:
Thanks for the update, then I say we merge this!

PR #174, opened 2019-10-29
Enable clippy in CI

m-ou-se:


korken89:
I think this issue ties a back to (and extends) our original discussion on enabling code style checks as well (can't find the issue right now...).
It's probably time to start that discussion again and take a decision. Comments @rust-embedded/all ?
Or should we first start with clippy and then expand to fmt checks?

m-ou-se:
The main reason I'd like Clippy in CI here, is so `#![deny(clippy::missing_inline_in_public_items)]` can be added, to prevent people from accidentally adding trivial functions without `#[inline]`. (E.g. #173)

thejpster:
Happy to start with clippy, and add `cargo fmt -- --check` separately, as that may require a reformat of the codebase.

thejpster:
Er, how do we find out why bors is sad?

m-ou-se:
Sometimes the `curl` command to download GCC from the ARM website fails. It was at the travis logs bors linked, before the build was restarted.

therealprof:
Build failed due to network error, I've retriggered a rebuild to make sure (unfortunately I hit the wrong button so it's not just retrying the one job that had failed).

therealprof:
bors retry

thejpster:
> 
> 
> Sometimes the `curl` command to download GCC from the ARM website fails. It was at the travis logs bors linked, before the build was restarted.

Hmm. Clicked all the little links and tiny buttons. Failed to see the great big blue link in the comment bors posted. Thanks!

PR #175, opened 2019-10-29
Enable the missing_inline_in_public_items clippy lint.

m-ou-se:
This adds `#![deny(clippy::missing_inline_in_public_items)]` to make sure all functions in this crate are marked `#[inline]`, unless they are explicitly marked with `#[allow(clippy::missing_inline_in_public_items)]`.

Only three functions in this crate are not `#[inline]`:

 - `write_words`
 - `write_all`
 - `write_aligned`

Additionally, the derived `Debug` impl's also have a non-inline implementations.
This unfortunately means that the allow attribute also needs to added to any types deriving `Debug`.

See also #171 and https://github.com/rust-embedded/cortex-m/pull/174#issuecomment-547304467.

### DIFF
--- a/src/itm.rs
+++ b/src/itm.rs
@@ -9,6 +9,7 @@ use aligned::{Aligned, A4};
 use crate::peripheral::itm::Stim;
 
 // NOTE assumes that `bytes` is 32-bit aligned
+#[allow(clippy::missing_inline_in_public_items)]
 unsafe fn write_words(stim: &mut Stim, bytes: &[u32]) {
     let mut p = bytes.as_ptr();
     for _ in 0..bytes.len() {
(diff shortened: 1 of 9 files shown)


thejpster:
Could we add a comment that explains under which circumstances it's appropriate to use the `#[allow(clippy::missing_inline_in_public_items)]` lint?

m-ou-se:
> Could we add a comment that explains under which circumstances it's appropriate to use the `#[allow(clippy::missing_inline_in_public_items)]` lint?

Done!

jonas-schievink:
> Additionally, the derived `Debug` impl's also have a non-inline implementations.
> This unfortunately means that the allow attribute also needs to added to any types deriving `Debug`.

This sounds like a bug in the lint – it's not really appropriate to warn about code the user doesn't control. I filed https://github.com/rust-lang/rust-clippy/issues/4861 to track this.

therealprof:
bors r=jonas-schievink

PR #180, opened 2019-12-18
Add Security Attribution Unit support

hug-dev:
The SAU is a Armv8-M core peripheral that, alongside the Implementation
Defined Attribution Unit, manages the security attribution of the memory
zones.
This driver provides abstraction to help setting the SAU up.

This pull-request adds a new dependency, `bitfield` which provides a really nice way to describe bit fields in registers with methods to access them. It makes it really easier to use/write rather than having to manually declare the shift number and mask constants.
Although `bitfield` does not have any dependency, I am well aware of the increased cost of memory its addition could cause to embedded targets.
I am really open to criticism about its use, if you would prefer me not to use it, or if you would like to see benchmarks of memory usage for some reference targets of embedded Rust 😃

### DIFF
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -70,7 +70,6 @@
 
 // TODO stand-alone registers: ICTR, ACTLR and STIR
 
-
 use core::marker::PhantomData;
 use core::ops;
 
@@ -90,6 +89,8 @@ pub mod fpu;
 pub mod itm;
 pub mod mpu;
 pub mod nvic;
+#[cfg(armv8m)]
+pub mod sau;
 pub mod scb;
 pub mod syst;
 #[cfg(not(armv6m))]
@@ -129,6 +130,9 @@ pub struct Peripherals {
 
     /// Nested Vector Interrupt Controller
     pub NVIC: NVIC,
+
+    /// Security Attribution Unit
+    pub SAU: SAU,
 
     /// System Control Block
     pub SCB: SCB,
@@ -189,6 +193,9 @@ impl Peripherals {
                 _marker: PhantomData,
             },
             NVIC: NVIC {
+                _marker: PhantomData,
+            },
+            SAU: SAU {
                 _marker: PhantomData,
             },
             SCB: SCB {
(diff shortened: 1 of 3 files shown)


src/peripheral/mod.rs
```diff
@@ -130,6 +131,9 @@ pub struct Peripherals {
     /// Nested Vector Interrupt Controller
     pub NVIC: NVIC,
 
+    /// Security Attribution Unit
+    pub SAU: SAU,
```
jonas-schievink: I think this should be `cfg`d out when not targeting armv8, and the doc comment should say that it only exists then.

I know we don't currently do this for other peripherals, but I think we should.

hug-dev: I was wondering the same! I thought there might be a reason for the fields not to be under `cfg` but only the `impl`?
If not I can do that change. Should I send another PR to apply `cfg` to all of the fields?

thejpster: > Should I send another PR to apply cfg to all of the fields?

I for one would like to see that.

thejpster:
> `bitfield` ... provides a really nice way to describe bit fields in registers

I think the medium-term plan was to find/write an SVD file for the Cortex-M core peripherals, so we could use svd2rust and generate a cortex-m-pac (which this high-level crate would then use). But I think this is a good interim step.

src/peripheral/mod.rs
hug-dev: Just made #181 as per this discussion 🚀

hug-dev: Will rebase on top of that one when (if) merged.

src/peripheral/mod.rs
```diff
@@ -130,6 +131,9 @@ pub struct Peripherals {
     /// Nested Vector Interrupt Controller
     pub NVIC: NVIC,
 
+    /// Security Attribution Unit
```
japaric: This is a breaking change. Adding a field to a structure that only has public fields is a breaking change.

"Why?" Because of destructuring syntax:

``` rust
fn foo(a: A) {
    let A { b, c } = a;
}

struct A {
    pub b: u8,
    pub c: u16,
    // pub d: u32, // adding this breaks the above `let` statement
}
```

hug-dev:
Maybe it is better to wait that the [cortex-m-pac](https://github.com/rust-embedded/cortex-m/issues/183#issuecomment-572762070) crate is available to remove the `bitfields` dependency and use that.

hug-dev:
Assuming that #193 is merged I guess this would not be a breaking change anymore and hence be merged more easily?
It could use the `bitfield` crate for now and I can switch to using the cortex-m-pac crate in a future PR when it is available.

src/peripheral/mod.rs
hug-dev: I assume this comment is now resolved because of the decision that has been taken during the meeting referenced in #181?

PR #187, opened 2020-01-12
0.6.2

### LINKED ISSUE #185: Publish 0.6.2
A few new features and improvements have landed that do not require a breaking change.

These changes still need changelog entries:
* https://github.com/rust-embedded/cortex-m/pull/173
* ~~https://github.com/rust-embedded/cortex-m/pull/167~~
* https://github.com/rust-embedded/cortex-m/pull/164

jonas-schievink:
Closes #185

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -25,7 +25,7 @@ fn main() {
     } else if target.starts_with("thumbv7em-") {
         println!("cargo:rustc-cfg=cortex_m");
         println!("cargo:rustc-cfg=armv7m");
-        //println!("cargo:rustc-cfg=armv7em");
+        println!("cargo:rustc-cfg=armv7em");  // (not currently used)
     } else if target.starts_with("thumbv8m.base") {
         println!("cargo:rustc-cfg=cortex_m");
         println!("cargo:rustc-cfg=armv8m");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.6.2] - 2020-01-12
+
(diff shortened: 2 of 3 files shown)


build.rs
```diff
@@ -19,12 +19,12 @@ fn main() {
     if target.starts_with("thumbv6m-") {
         println!("cargo:rustc-cfg=cortex_m");
         println!("cargo:rustc-cfg=armv6m");
-    } else if target.starts_with("thumbv7m-") {
-        println!("cargo:rustc-cfg=cortex_m");
-        println!("cargo:rustc-cfg=armv7m");
-    } else if target.starts_with("thumbv7em-") {
+    } else if target.starts_with("thumbv7m-") || target.starts_with("thumbv7em-") {
```
therealprof: I'd rather we do not make this change. If the compiler happens to add special support for `thumbv7em` which is more than just the DSP extension, e.g. saturated instructions and FPU support, then we do not need to update this crate and all dependencies again.

jonas-schievink: This is not a functional change, it just silences a Clippy lint which otherwise fails CI on nightly. Unless I misunderstand what you're saying?

therealprof: I'm refererring to the comment new comment below. This code changes the flags passed to rustc based on the architecture we're compiling for and you're saying it does not make sense to pass the `armv7em` flag because it's not supported at the moment.

If this fixes a clippy lint (**not** mentioned in the comment) and has not other drawbacks (which **is** implied by the comment, namely that we'll have to add it back once supported) I'm okay with the change but would like to see those facts recorded in the comment. To me it currently reads as if we're simply removing this because we don't need to have now but will later and then have to add it back.

therealprof:
... and published.

Issue #184, opened 2020-01-08
Consider getting rid of aligned dependency subtree


nickray:
This eventually pulls in two versions of generic-array which is a bit weird. Maybe take over `aligned` and/or `as-slice` and add some feature flags there? Or make `itm::write_aligned` itself an optional feature?

re. https://github.com/rust-embedded/cortex-m/issues/183

jonas-schievink:
Oh *this* is where all the `generic-array` dependencies come from? I was wondering about that.

Pulling in 4 dependencies for a single function seems very excessive, so I'm on board with removing them. We can add our own alignment-enforcing struct to the `itm` module instead.

nickray:
Yeah for some (compatibility?) reason as-slice pulls in both.

therealprof:
Yeah, otherwise you run into weird compatibility problems.

nickray:
None of my dependencies that depend on generic-array use 0.12, so I' be curious what compatibility issues there might be?

PR #190, opened 2020-01-15
Remove `aligned`, improve ITM code

jonas-schievink:
This does a handful of things:

* Removes the dependency on `aligned` (and thus `as-slice` and the 2 versions of `generic-array`), instead providing our own simplified `Aligned` wrapper
* Moves the innards of `write_aligned` to its own function, and calls that from `write_all` instead of transmuting `&[u8]` to `&Aligned<A4, [u8]>` (which is likely UB)
* Fixes the doc example, which didn't compile anymore
* Sinks the `#[allow]` attributes into the functions so they only cover the statements they need to

Closes https://github.com/rust-embedded/cortex-m/issues/184

### DIFF
--- a/src/itm.rs
+++ b/src/itm.rs
@@ -1,21 +1,57 @@
 //! Instrumentation Trace Macrocell
 //!
-//! **NOTE** This module is only available on ARMv7-M and newer
+//! **NOTE** This module is only available on ARMv7-M and newer.
 
-use core::{fmt, mem, ptr, slice};
-
-use aligned::{Aligned, A4};
+use core::{fmt, ptr, slice};
 
 use crate::peripheral::itm::Stim;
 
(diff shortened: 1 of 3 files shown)


src/itm.rs
```diff
@@ -19,6 +16,45 @@ unsafe fn write_words(stim: &mut Stim, bytes: &[u32]) {
     }
 }
 
+/// Writes an aligned byte slice to the ITM.
+///
+/// `bytes` must be 4-byte aligned.
```
adamgreig: Should this be `buffer`?

jonas-schievink: Yes, good catch!

adamgreig:
Worth noting that this is a breaking change due to the different `Aligned`.

therealprof:
Nice!

Issue #188, opened 2020-01-13
"Invalidate Cache" operation is unsafe


jonas-schievink:
The optional Cortex-M cache exposes "clean" and "invalidate" operations. As pointed out by @adamgreig in https://github.com/rust-embedded/cortex-m/issues/47#issuecomment-573482398, performing only the "invalidate" operation without a preceding "clean" is unsound.

* "Clean" flushes cache lines to main memory and makes operations visible to other peripherals and cores
* "Invalidate" drops cache lines so that they are fetched from main memory on subsequent accesses, making modifications from other bus masters visible to the core doing the invalidation

This operation is currently exposed via safe interfaces:
* As methods on [`CBP`](https://docs.rs/cortex-m/0.6.2/cortex_m/peripheral/struct.CBP.html)
* As methods on [`SCB`](https://docs.rs/cortex-m/0.6.2/cortex_m/peripheral/struct.SCB.html)

We should either deprecate the methods and add unsafe replacements, or make the methods `unsafe` in a breaking change without deprecation period (this avoids having to find new names).

In my opinion there is little to no value in exposing only "invalidate" since it hard to use correctly and has little to no benefits over "clean + invalidate", so removing it is also an option.

adamgreig:
> performing only the "invalidate" operation without a preceding "clean" is unsound

It might also be that a clean is unsound too, I want to think about this a little more.


> make the methods unsafe in a breaking change without deprecation period

Given the planned upcoming breaking changes and I think the relatively uncommon use of the cache operations this is my preferred option.

> In my opinion there is little to no value in exposing only "invalidate" since it hard to use correctly and has little to no benefits over "clean + invalidate", so removing it is also an option.

These operations are peripheral registers in Cortex-M, with clean+invalidate being distinct from invalidate - sometimes you need one and sometimes you need the other. The methods in SCB mirror those provided in CMSIS. You very often need to invalidate without cleaning if you know another observer has written to the main memory and your cache is outdated, since a clean+invalidate would overwrite whatever the observer had written. Clean+invalidate is pretty much just used for disabling the cache, when you want to ensure all updates are written out and then the item is removed from the cache. I definitely think we should keep all the methods.

jonas-schievink:
Fixed by https://github.com/rust-embedded/cortex-m/pull/192, closing

PR #192, opened 2020-01-15
Update and improve cache operations.

### LINKED ISSUE #47: RFC: Even higher-level cache maintenance operations
#40 landed CMSIS-style DCache cleaning and invalidating by address and size, both `usize`. This maps well to the actual hardware operations but is inconvenient to use.

We could add some convenience methods that operate on normal Rust objects:

* `clean_dcache_by_slice<T>(slice: &[T])` which passes `slice.as_ptr() as usize` and `core::mem::size_of<T>() * slice.len()` to the underlying methods,
* `clean_dcache_by_ref<T>(obj: &T)`, which passes `T as *const _ as usize` and `core::mem::size_of<T>()` to the underlying methods.
* As above for `invalidate_dcache` and `clean_invalidate_dcache`

Big safety issue: all the cache methods operate on cache lines of 32 bytes. If you pass something whose size is not a multiple of 32, then some extra bytes after that object will be cleaned or invalidated. In general this is definitely unsafe. Perhaps these methods could panic on non-32-byte-multiple sizes?

adamgreig:
Closes #47, #188.

I've implemented the proposed methods from #47 and marked all d-cache invalidation functions as unsafe. It's not unsafe to invalidate i-cache or branch predictor as they are read-only caches. The clean and clean+invalidate operations do not alter memory from the executing core's point of view so are also safe.

It wasn't possible to remove the requirement to pass in `&mut CPUID` as you require synchronized access to `CPUID` to read the number of sets and ways in the cache, which is required to fully clean or invalidate them, which is required to enable or disable them. So it goes.

Breaking change due to changing safety of d-cache invalidation functions.

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -314,7 +314,7 @@ use self::scb_consts::*;
 
 #[cfg(not(armv6m))]
 impl SCB {
-    /// Enables I-cache if currently disabled
+    /// Enables I-cache if currently disabled.
     ///
     /// This operation first invalidates the entire I-cache.
     #[inline]
@@ -338,7 +338,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Disables I-cache if currently enabled
+    /// Disables I-cache if currently enabled.
     ///
     /// This operation invalidates the entire I-cache after disabling.
     #[inline]
@@ -362,7 +362,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Returns whether the I-cache is currently enabled
+    /// Returns whether the I-cache is currently enabled.
     #[inline(always)]
     pub fn icache_enabled() -> bool {
         crate::asm::dsb();
@@ -372,7 +372,7 @@ impl SCB {
         unsafe { (*Self::ptr()).ccr.read() & SCB_CCR_IC_MASK == SCB_CCR_IC_MASK }
     }
 
-    /// Invalidates entire I-cache
+    /// Invalidates the entire I-cache.
     #[inline]
     pub fn invalidate_icache(&mut self) {
         // NOTE(unsafe): No races as all CBP registers are write-only and stateless
@@ -385,7 +385,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Enables D-cache if currently disabled
+    /// Enables D-cache if currently disabled.
     ///
     /// This operation first invalidates the entire D-cache, ensuring it does
     /// not contain stale values before being enabled.
@@ -407,7 +407,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Disables D-cache if currently enabled
+    /// Disables D-cache if currently enabled.
     ///
     /// This operation subsequently cleans and invalidates the entire D-cache,
     /// ensuring all contents are safely written back to main memory after disabling.
@@ -426,7 +426,7 @@ impl SCB {
         self.clean_invalidate_dcache(cpuid);
     }
 
-    /// Returns whether the D-cache is currently enabled
+    /// Returns whether the D-cache is currently enabled.
     #[inline]
     pub fn dcache_enabled() -> bool {
         crate::asm::dsb();
@@ -436,7 +436,7 @@ impl SCB {
         unsafe { (*Self::ptr()).ccr.read() & SCB_CCR_DC_MASK == SCB_CCR_DC_MASK }
     }
 
-    /// Invalidates entire D-cache
+    /// Invalidates the entire D-cache.
     ///
     /// Note that calling this while the dcache is enabled will probably wipe out the
     /// stack, depending on optimisations, therefore breaking returning to the call point.
@@ -461,7 +461,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Cleans entire D-cache
+    /// Cleans the entire D-cache.
     ///
     /// This function causes everything in the D-cache to be written back to main memory,
     /// overwriting whatever is already there.
@@ -483,7 +483,7 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Cleans and invalidates entire D-cache
+    /// Cleans and invalidates the entire D-cache.
     ///
     /// This function causes everything in the D-cache to be written back to main memory,
     /// and then marks the entire D-cache as invalid, causing future reads to first fetch
@@ -506,10 +506,10 @@ impl SCB {
         crate::asm::isb();
     }
 
-    /// Invalidates D-cache by address
+    /// Invalidates D-cache by address.
     ///
-    /// * `addr`: the address to invalidate, which must be cache-line aligned
-    /// * `size`: number of bytes to invalidate, which must be a multiple of the cache line size
+    /// * `addr`: The address to invalidate, which must be cache-line aligned.
+    /// * `size`: Number of bytes to invalidate, which must be a multiple of the cache line size.
     ///
     /// Invalidates D-cache cache lines, starting from the first line containing `addr`,
     /// finishing once at least `size` bytes have been invalidated.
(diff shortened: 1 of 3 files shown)


src/peripheral/scb.rs
```diff
@@ -518,21 +555,80 @@ impl SCB {
...
+    /// Invalidation causes the next read access to memory to be fetched from main memory instead
+    /// of the cache.
+    ///
+    /// # Safety
+    ///
+    /// After invalidating, `obj` will be read from main memory on next access. This may cause
+    /// recent writes to `obj` to be lost, potentially including the write that initialised it.
+    /// Therefore, this method may cause uninitialised memory or invalid values to be read,
+    /// resulting in undefined behaviour. You must ensure that main memory contains a valid and
+    /// initialised value for T before invalidating `obj`.
```
jonas-schievink: Isn't it unsound in general when the memory a `&T` points to is modified (unless `T` is an `UnsafeCell`)?

adamgreig: Do you think we should require `&mut` here instead of `&`, or just give a more dire warning about potential consequences?

jonas-schievink: `&mut` sounds like a good option, as long as it doesn't make the API unusable. Do you happen to have any real-world examples where this API would be used?

adamgreig: I've used the previous invalidate API in ethernet peripheral drivers where the descriptors and packet memory are shared between CPU and peripheral. In all of those instances the code had &mut access, though the receive buffers were in principle read-only.

The other use case I can see is sharing memory between cores on a multi-core device, e.g. the STM32H7 dual-core series have D-cache for the M7 core, so communicating with the M4 core could require cache management. Imagine a synchronisation primitive which needs to invalidate some interior memory to ensure it's reading a fresh value from the other core, and wants to be able to do this with only `&self`. But probably such a primitive should contain an `UnsafeCell` which can handle the interior mutability, and then it would still be able to pass &mut to the cache invalidation function.

On balance I think it's probably a more obvious API if it takes `&mut` to make it clear that the value might change, and in the rare case someone needs to access it with a non-mut, it's only a one line function that calls another public unsafe method anyway, so it's not that much extra effort. I'll update this PR.

adamgreig:
A couple of other notes:

* The comments refer to 32-byte alignment and length. Technically in ARMv7-M this length is core-specific and available from the DMinLine field of the CTR register. Currently, only Cortex-M7 cores can have a D-cache, and Cortex-M7 cores can only have 32-byte DMinLine. In the future ARM might release other cores with other cache sizes. I'll update the docstrings to clarify.
* The above means we probably should not have an assert that the memory being invalidated is 32-byte aligned and a multiple of 32 bytes long. Which is a shame as those asserts would probably be quire useful and make this a less sharp API. Perhaps we could read the DMinLine register and check based on that, I'll investigate.

Please don't merge just yet.

adamgreig:
I've updated the documentation to more clearly convey what's required and what will happen, updated the methods to take `&mut` where appropriate, and updated the underlying loops to use `DminLine` to obtain the cache size at runtime instead of hardcoding 32, although at the moment all released devices will use 32.

This is ready for review now.

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    /// easily leading to memory corruption and undefined behaviour. This precondition is checked
+    /// in debug builds using a `debug_assert!()`, but not checked in release builds to avoid
+    /// a runtime-dependent `panic!()` call.
+    #[inline]
+    pub unsafe fn invalidate_dcache_by_slice<T>(&mut self, slice: &mut [T]) {
+        self.invalidate_dcache_by_address(slice.as_ptr() as usize,
+                                          slice.len() * core::mem::size_of::<T>());
+    }
+
     /// Cleans D-cache by address
```
jonas-schievink:
```suggestion
    /// Cleans D-cache by address.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
     ///
-    /// `addr`: the address to clean
-    /// `size`: size of the memory block, in number of bytes
+    /// * `addr`: the address to clean
```
jonas-schievink:
```suggestion
    /// * `addr`: The address to start cleaning at.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    /// * `size`: number of bytes to clean
```
jonas-schievink:
```suggestion
    /// * `size`: The number of bytes to clean.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
             cbp.dccmvac(addr as u32);
-            addr += LINESIZE;
+            addr += line_size;
         }
 
         crate::asm::dsb();
         crate::asm::isb();
     }
 
+    /// Cleans an object in D-cache
```
jonas-schievink:
```suggestion
    /// Cleans an object from the D-cache.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    ///
+    /// * `obj`: Object to clean
```
jonas-schievink:
```suggestion
    /// * `obj`: The object to clean.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    /// It is recommended that `obj` is both aligned to the cache line size and a multiple of
+    /// the cache line size long, otherwise surrounding data will also be cleaned.
+    ///
+    /// Cleaning the cache causes whatever data is present in the cache to be immediately written
+    /// to main memory, overwriting whatever was in main memory.
+    pub fn clean_dcache_by_ref<T>(&mut self, obj: &T) {
+        self.clean_dcache_by_address(obj as *const T as usize, core::mem::size_of::<T>());
+    }
+
+    /// Cleans a slice in D-cache
```
jonas-schievink:
```suggestion
    /// Cleans a slice from D-cache.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    ///
+    /// * `slice`: Slice to clean
```
jonas-schievink:
```suggestion
    /// * `slice`: The slice to clean.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    /// the cache line size long, otherwise surrounding data will also be cleaned.
+    ///
+    /// Cleaning the cache causes whatever data is present in the cache to be immediately written
+    /// to main memory, overwriting whatever was in main memory.
+    pub fn clean_dcache_by_slice<T>(&mut self, slice: &[T]) {
+        self.clean_dcache_by_address(slice.as_ptr() as usize,
+                                     slice.len() * core::mem::size_of::<T>());
+    }
+
     /// Cleans and invalidates D-cache by address
```
jonas-schievink:
```suggestion
    /// Cleans and invalidates D-cache by address.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
     ///
-    /// `addr`: the address to clean and invalidate
-    /// `size`: size of the memory block, in number of bytes
+    /// * `addr`: the address to clean and invalidate
```
jonas-schievink:
```suggestion
    /// * `addr`: The address to clean and invalidate.
```

src/peripheral/scb.rs
```diff
@@ -486,88 +508,254 @@ impl SCB {
...
+    /// * `size`: number of bytes to clean and invalidate
```
jonas-schievink:
```suggestion
    /// * `size`: The number of bytes to clean and invalidate.
```

adamgreig:
Thanks, I've applied suggestions and updated some other `initialised` :gb:

jonas-schievink:
Great, let's do this!

bors r+

jonas-schievink:
Apparently we're downloading stuff from Arm servers during CI. Not sure why but we shouldn't be doing that.

adamgreig:
bors retry

>

adamgreig:
We download the gcc toolchain which is used to verify the precompiled
binary blobs are correct compared to the source, iirc.

jonas-schievink:
Using `apt-get` is better for that since Travis caches it. I'll open an issue.

adamgreig:
At the time I don't think a suitable version was available in apt and it
still might not be. We could have Travis cache the downloaded toolchain,
though.

PR #193, opened 2020-01-16
Make `Peripherals` non-exhaustive and improve its docs

jonas-schievink:
This means that it's no longer a breaking change to add fields to it, which is important since Arm is likely to add more in upcoming architectures. They could also add extensions that add peripherals.

### DIFF
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -89,7 +89,6 @@ mod test;
 
 /// Core peripherals
 #[allow(non_snake_case)]
-#[non_exhaustive]
 pub struct Peripherals {
     /// Cache and branch predictor maintenance operations (not present on Cortex-M0 variants)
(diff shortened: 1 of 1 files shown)


hug-dev:
I am in favor of this as it would probably remove the `breaking-change` label on #180!

therealprof:
Build failure is unrelated to the change but I guess we need to address this:
```
error: unnecessary parentheses around assigned value
269   --> src/peripheral/scb.rs:308:38
270    |
271308 |     pub const SCB_CCR_IC_MASK: u32 = (1 << 17);
272    |                                      ^^^^^^^^^ help: remove these parentheses
273    |
274    = note: `-D unused-parens` implied by `-D warnings`
```

thejpster:
bors retry

jonas-schievink:
bors retry

PR #201, opened 2020-03-15
Add UDF instruction. Closes #199.

### LINKED ISSUE #199: Expose the `udf` instruction
It's useful as a stable replacement for `core::intrinsics::abort`.

adamgreig:


jonas-schievink:
I wonder if we should make the function diverge. There is not really any useful recovery that an exception handler could perform (and such a handler can't currently be written anyways).

adamgreig:
It's not necessarily _useful_ but if you defined a custom `UsageFault` handler (which you can do in c-m-rt) which did nothing, would execution continue from after the `udf` instruction, or do you immediately UsageFault again?

I think irregardless if you're deliberately calling `cortex_m::asm::udf()` you probably don't intend to continue execution from this branch, so I'll update the call to be divergent and add an infinite loop.

jonas-schievink:
A `UsageFault` handler that deliberately continues past opcodes that are expected to trap is certainly UB, so we don't really have to worry about that. This means that it should be fine to assume that the `udf` itself diverges, so the `bx lr` can be omitted and the infinite loop can be replaced with `unreachable_unchecked`.

In some cases fault handlers can return safely, for example when a userspace process causes a `HardFault` (the kernel might terminate the process and schedule another), but that's not really an issue here (though `udf` does cause a HardFault on thumbv6 from what I recall).

adamgreig:
Sure, makes sense. I've updated to use `unreachable_unchecked()`.

PR #202, opened 2020-03-17
Prevent unnecessary bounds check in SCB::{get_priority, set_priority}

qwerty19106:
SystemHandler.index() gives true index for SCB.shpr array.
But now it produce unnecessary bounds check. For example, SCB::set_priority:
```
48: sym.cortex_m::peripheral::scb::__impl_cortex_m::peripheral::SCB_::set_priority::h622cd21bcc3321be ();
0x08000376      push    {r7, lr}
0x08000378      mov     r7, sp
0x0800037a      bl      cortex_m::peripheral::scb::SystemHandler::index::hdaf1a31aa5a6a8c5 ; sym.cortex_m::peripheral::scb::SystemHandler::index::hdaf1a31aa5a6a8c5
0x0800037e      subs    r0, 4
0x08000380      uxtb    r0, r0
0x08000382      cmp     r0, 0xb    ; 11
0x08000384      itttt   ls
0x08000386      movw    r1, 0xed18
0x0800038a      movt    r1, 0xe000
0x0800038e      movs    r2, 0xff
0x08000390      strb    r2, [r0, r1]
0x08000392      it      ls
0x08000394      pop     {r7, pc}
0x08000396      movw    r2, 0x1dc0
0x0800039a      movs    r1, 0xc
0x0800039c      movt    r2, 0x800
0x080003a0      bl      core::panicking::panic_bounds_check::h5181f47ae17a04a5 ; sym.core::panicking::panic_bounds_check::h5181f47ae17a04a5
0x080003a4      trap
```

This PR fix it.

PR #203, opened 2020-03-17
Use u8 repr for enum instead of custom method

therealprof:
This is an alternative proposal to #202 as suggested in
https://github.com/rust-embedded/cortex-m/pull/202#issuecomment-600017553


therealprof:
Not sure, that's what I want to find to (and also whether this compiles with the MSRV), but it certainly is at least an improvement for `dev` builds because the `function` for some reason wasn't even regular `#[inline]`.

jonas-schievink:
This is a 1.0 feature, so the only concern is codegen.

I do believe rustc will attach LLVM range metadata, so it should optimize well.

therealprof:
Well, it did build, even with the current MSRV. And it definitely is an improvement in codegen for dev builds at least and most certainly no regression for release builds. 😉

PR #209, opened 2020-04-21
Added arch specific mutex implementation

korken89:
I started the implementation for an arch specific mutex to get discussion going.


src/mutex.rs
```diff
@@ -0,0 +1,25 @@
+//! Implementation of a critical section based mutex that also implements the `mutex-trait`.
+
+use core::cell::RefCell;
+
+/// A critical section based mutex
```
jonas-schievink, line 5:
```suggestion
/// A critical section based mutex.
```

jamesmunns:
Hey @korken89, I'm not sure it's possible to share this mutex between an interrupt and main with this, at least without using `unsafe`, as your mutex is not Sync.

Additionally, it's kind of a bummer that this mutex panics if the resource is taken, but I guess that was already decided based on the Mutex-Trait API.

korken89:
Right, I have to move to `UnsafeCell`

jonas-schievink:
What happens now if I lock the mutex while already holding it?

korken89:
Ah right, now I remember the original issue around the implementation - thanks for pointing it out.
I'll take a another look the upcoming days.

therealprof:
Discussion from the meeting: Seems this has stalled out. Maybe this could be another focus project? If people are interested in joining such a group, please speak up.

The MSP430 team is considering taking over the current implementation from bare-metal to not keep unneeded dependencies around.

newAM:
I'm triaging Cortex-M issues, I think this has been solved with [critical-section](https://github.com/rust-embedded/critical-section).  Feel free to re-open if there's still something to discus here!

PR #210, opened 2020-04-28
Use llmv_asm! macro for inline assembly

### LINKED ISSUE #204: Inline assembly should switch to `llvm_asm!`
`asm!` will be deprecated soon https://github.com/rust-lang/rust/pull/71007

stianeklund:
Fixes #204

the `asm!` macro will soon be deprecated.
This changes all of the `asm!` calls to `llvm_asm!`

To my knowledge doing a direct replacement should be fine.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(llvm_asm)]
 //! Low level access to Cortex-M processors
 //!
 //! This crate provides:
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(llvm_asm)]
```
japaric, line 1: `#![feature(..)]` is not allowed in the stable channel so this should also be behind the "inline-asm" Cargo feature. You'll need to write something like this:

``` rust
#![cfg_attr(feature = "inline-asm", feature(llvm_asm))]
```

Also you should move this to after the crate level doc comment so it sits next to the other crate level attributes.

stianeklund: Ah my bad. I made a mistake here, I thought I had compiled this with the stable toolchain before committing, turns out I did not.

PR #212, opened 2020-04-28
Add symtab (size info) for all external assembly functions

### LINKED ISSUE #211: add (.symtab) size info to external assembly functions
Our external assembly subroutines (e.g. `__nop`) are not properly reporting their size in the .symtab section. If you build the cortex-m-quickstart template and print the contents of the symbol table (.symtab) you'll see that those assembly subroutines report their size as 0.

``` console
$ arm-none-eabi-readelf -s target/thumbv7m-none-eabi/release/app
(..)
     8: 0000040b    10 FUNC    LOCAL  DEFAULT    2 _ZN3app18__cortex_m_rt_ma
(..)
    21: 00000415   154 FUNC    GLOBAL DEFAULT    2 Reset
(..)
    30: 000004b3     0 FUNC    GLOBAL DEFAULT    2 __nop
```

The first two lines (Rust functions) are correct. The last line (`__nop`) is wrongly reporting its size as 0.

This bug won't affect linking or the runtime behavior of the program but may cause ELF post-processing tools to incorrectly interpret these functions as tags (e.g. `$t.0`), which are also zero sized, and discard them which may result in invalid output binaries.

To add size info you need to add the following line to the existing subroutines in the `asm*.s` files:

``` diff
   .section .text.__bkpt
   .global __bkpt
   .thumb_func
 __bkpt:
   bkpt
   bx lr
+  .size __bkpt, . - __bkpt
```

Note that the name of the subroutine needs to be used in the `.size` directive (`__bkpt` in the example)

After modifying those files you'll need to re-assemble the files by running the `./assemble.sh` script. 

After making those changes you should see the correct size in the output of `readelf`:

``` console
$ arm-none-eabi-readelf -s target/thumbv7m-none-eabi/release/app
(..)
    30: 000004b3     4 FUNC    GLOBAL DEFAULT    2 __nop
```

P.S. many other crates that use external assembly (cortex-m-rt, cortex-m-semihosting, etc.) need this fix

stianeklund:
Adds `.size` info for all of the external assembly functions. 
Fixes #211


jonas-schievink:
Don't these have to go after the implementation? They use `.`, which refers to the current address.

stianeklund:
> Don't these have to go after the implementation? They use `.`, which refers to the current address.
@jonas-schievink Yeah I think  you're right. I've made the changes.

jonas-schievink:
bors r-

CI broke

japaric:
Most of the CI errors should be fixed by #210 so this PR is blocked on that one.

This CI error makes sense no sense to me though: https://travis-ci.org/github/rust-embedded/cortex-m/jobs/680770633

``` console
+diff -u bin/thumbv6m-none-eabi.after bin/thumbv6m-none-eabi.before
--- bin/thumbv6m-none-eabi.after	2020-04-28 22:06:50.010978643 +0000
+++ bin/thumbv6m-none-eabi.before	2020-04-28 22:06:49.911028639 +0000
@@ -1,6 +1,6 @@
 In archive bin/thumbv6m-none-eabi.a:
-cortex-m.o:     file format elf32-littlearm
+$crate.o:     file format elf32-littlearm
 Disassembly of section .text.__bkpt:
```

Why is there a `$crate.o` in the output? Did the variable expansion in `assemble.sh` fail somehow?

jonas-schievink:
@japaric I think that was some weirdness with Travis' bash version, it should be fixed now

jonas-schievink:
@stianeklund please rebase on master and rebuild the archives, then we can land this

PR #213, opened 2020-05-03
Generate debug information for assembly functions

Tiwalun:
This helps with debugging, but makes the binaries a bit larger.


adamgreig:
To be clear: this makes our object files larger, but shouldn't have any impact on application code size, right?

Tiwalun:
It makes the final ELF files bigger, the debug information for the assembly function will always be there, regardless of the rustc settings.

The actual code size without debug information is unchanged. For an embedded context, this is probably more relevant.

adamgreig:
Sorry, yes, I mean "the final occupied space in the target flash" is unchanged, whereas we don't really care if the elfs are a few bytes larger.

adamgreig:
We'll have to wait for #214 which I thought would merge before this.

adamgreig:
bors retry

adamgreig:
bors retry

PR #219, opened 2020-05-25
ITM: don't test reserved bits in is_fifo_ready

cbiffle:
On ARMv7-M, bits 31:1 of the value read from STIMx are reserved, so
comparing them against zero is a bad idea.

On ARMv8-M, bit 1 has been repurposed to indicate DISABLED. This means
that the is_fifo_ready impl hangs forever when ITM is disabled on a
Cortex-M33 (for example).

Changed to test only the FIFOREADY bit.

@bcantrill

### DIFF
--- a/src/peripheral/itm.rs
+++ b/src/peripheral/itm.rs
@@ -55,6 +55,6 @@ impl Stim {
     /// Returns `true` if the stimulus port is ready to accept more data
     #[inline]
     pub fn is_fifo_ready(&self) -> bool {
-        unsafe { ptr::read_volatile(self.register.get()) == 1 }
+        unsafe { ptr::read_volatile(self.register.get()) & 1 == 1 }
     }
 }


cbiffle:
Ugh. So, I have bad news.

On the one hand, this change makes v7-M more correct.

On the other hand, it turns out that this isn't sufficient to fix v8-M. Here's the deal.

- Bit 1 has been repurposed to mean DISABLED on v8-M.
- Bit 0 is still FIFOREADY
- The reset state appears to be DISABLED and (importantly) not FIFOREADY.
- On v7M the reset state is undefined, but on the STM32F407 at least it is FIFOREADY.

So, on reset, with no debugger attached, if one wants ITM writes to succeed, you've got two options:

1. Implement architecture-version-specific logic.
2. Read the reserved bits I just fixed. :-)

While ARM described the bits on ARMv7-M as reserved, if you look at [their pseudocode](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0337e/ch11s06s01.html), they're treating anything other than 0 as ready to write. This logic works on both v7-M and v8-M but is _technically_ dealing in reserved bits.

I'm happy to fix this either way, but I wonder if anyone has a strong opinion on approach?

adamgreig:
Ah, that's a nuisance! I think architecture-specific logic is the way to go then. Keep the current behaviour for thumbv7 and add the alternative check of DISABLED and FIFOREADY for thumbv8. We already do this sort of check extensively in this crate so I don't see any reason not to do it here.

PR #220, opened 2020-05-27
Implement accessing FPSCR

bugadani:
On the nRF52, sometimes it is necessary to manipulate the FPSCR register, otherwise the device wakes up immediately from sleep. (At least on this device) the FPSCR is only available through `vmrs` instructions.

I've implemented reading the register, parsing its bits and writing a raw value to the register, but let me know if I should also implement manipulation of the named bits.

I would also like to request some assistance to get this to actually build, it's not clear to me how `.s` files are compiled in this library. I'm also not certain where the actual place for this would be - in the registers, or in the `fpu` module.


adamgreig:
At first glance this looks like a great addition, thanks! I think it's fine in `registers` as it's not part of the FPU memory-mapped space, it's a CPU register.

For building the assembly file, we do that manually to include the object file in the commit, and the CI system does it again to verify that the object file matches the assembly source. That's done in [check_blobs.py](https://github.com/rust-embedded/cortex-m/blob/master/check-blobs.sh) and you should be able to follow that to build it yourself too.

As for manipulation of named bits, I think just having the ability to read them and write raw values is a good start and adds a feature we don't support at the moment, but if you did want to implement setting bits it could be done similar to the current `control` register.

bugadani:
Well, it's probably 15 commits more than it should have been, but why hide my mistakes. I believe this is ready for review, however, I'm not able to test this on armv8.

PR #226, opened 2020-06-09
Expose the orphaned ICTR/ACTLR registers.

cbiffle:
In ARMv7-M these were floating alone in system control space, without a
unifying name. As a result, they weren't exposed in earlier versions of
this crate.

In ARMv8-M they have been given a name, the Implementation Control
Block, and more registers have been added. I've used that name for all
architecture revisions.


cbiffle:
That Travis failure looks like flake (your build is downloading things from ARM and failed). It's not immediately obvious to me how to retrigger it.

Issue #232, opened 2020-07-02
`enable_icache` is wrong on M7 without `inline-asm` feature


cbiffle:
This has been a fun one to track down.

**tl;dr:** if enabling instruction and/or data caches on your M7 is causing weird crashes, turn on the `cortex_m` `inline-asm` feature and make sure you're compiling at `--release`. Oh, and you're on nightly now. Sorry.

# Details

I've got an STM32H7B3 here, and shortly after enabling the icache I would _sometimes_ get hard faults that looked suspiciously like it trying to execute random data. (I'm doing this early enough in boot that I don't have configurable faults broken out, but they were either BF or MMF.)

The key here is _sometimes_ -- some builds worked okay. Until I tried turning on dcache. Then things went squirrely.

It came down to: _how many instructions did the compiler decide to insert between turning on the cache and the `dsb` / `isb` sequence?_ If the answer is not zero and contains a branch or memory access, _weird things_ will happen.

The current implementation of `enable_icache` reads as follows:
```rust
    #[inline]
    pub fn enable_icache(&mut self) {
        // Don't do anything if I-cache is already enabled
        if Self::icache_enabled() {
            return;
        }

        // NOTE(unsafe): No races as all CBP registers are write-only and stateless
        let mut cbp = unsafe { CBP::new() };

        // Invalidate I-cache
        cbp.iciallu();

        // Enable I-cache
        // NOTE(unsafe): We have synchronised access by &mut self
        unsafe { self.ccr.modify(|r| r | SCB_CCR_IC_MASK) };

        crate::asm::dsb();
        crate::asm::isb();
    }
```

The tail end of that function is _almost_ right.

- The `modify` will end in a `str`.
- The barriers will follow.

However,

- There's no use of `compiler_fence` or equivalent to prevent instructions from being rearranged above the barriers. (Perhaps we can assume this won't happen because they're `extern "C"`? I'm _fairly certain_ I just saw it happen but I didn't save the listing.)
- `dsb` and `isb` _default to being function calls_

This means the actual dynamic instruction trace here is

```
str r1, [r0, #0]
bl __dsb
dsb sy
bx lr
bl __isb
isb sy
bx lr
```

In other words, we've got one branch in the pipeline before the write even takes effect, and no fewer than three before the `isb` causes a prefetch/pipeline flush. These instructions often execute incorrectly. (On a reasonably fast M7, "often" becomes "always" in my testing.)

[...]

PR #234, opened 2020-07-04
Use assembly sequences to enable caches.

cbiffle:
See #232, which this partially fixes -- there's still the question of
taking an interrupt in the midst of these sequences.


cbiffle:
I'm not clear on why `as` on one of your builders has stronger opinions about the use of unified thumb mnemonics than the others, but I have removed the offending suffix.

adamgreig:
Only the x86 target runs the `assemble.sh` script to check the pre-built blobs match the CI-generated ones, the other targets just test the crate itself.

Our CI is using ARM's `gcc-arm-none-eabi-7-2018-q2-update-linux` which is a bit old; what version are you assembling with?

cbiffle:
I am using binutils 2.34-1.

cbiffle:
Is the decision to use pre-unified syntax in the asm files deliberate? It means that the asm files and the inline-asm blocks use different syntax. (I'm having to go digging to remember how to express a reg-reg-immed orr in pre-unified mnemonics.)

If not, I will add the `.syntax unified` directive to the top of the file.

adamgreig:
I don't see any reason not to use unified syntax here, I assume it was an oversight originally.

The replication between the asm files and the inline asm blocks is the only way we've found to provide asm to stable users while still offering inline asm for nightly users, but I don't think there's any need for inline asm for these cache management operations, so we could only ship the prebuilt objects here. If we end up including all the other maintenance operations I think there would be enough assembly to tip the balance towards only using the prebuilt objects; as-is it's short enough that I think we might as well include both. Hopefully soon Rust will include stable inline assembly and eventually we can remove the duplication.

On the critical section: do you think there's any situation where users are likely to be worried about the overhead of entering a new critical section, justifying a version of the function that takes an `&CriticalSection` token? I assume most users are not enabling the cache in their fast paths, and as you've noted unprivileged users won't be changing the cache either. If that's the case I think just stick the `cps` inside the assembly and document it in the function docstring.

cbiffle:
> I don't see any reason not to use unified syntax here, I assume it was an oversight originally.

woot, updated

> The replication between the asm files and the inline asm blocks is the only way we've found to provide asm to stable users while still offering inline asm for nightly users

Right. I'd do the same thing in your place. I wish that `llvm_asm!` could be pointed at a file, though, so that they both eat the same input. It's not immediately clear that that's possible (without, say, a proc macro).

> I don't think there's any need for inline asm for these cache management operations

Well, that simplifies things! :-)

> On the critical section: do you think there's any situation where users are likely to be worried about the overhead of entering a new critical section, justifying a version of the function that takes an &CriticalSection token?

This may be a failure of imagination, but...no, I don't think that's likely. I haven't quantified the overhead on the M7 (which is where caches start to become commonplace) but I would be surprised if anyone noticed. So I will add the instructions and make the code correct.

src/peripheral/scb.rs
```diff
@@ -400,11 +406,17 @@ impl SCB {
...
-        unsafe { self.ccr.modify(|r| r | SCB_CCR_DC_MASK) };
+        extern "C" {
+            // see asm-v7m.s
+            fn __enable_dcache();
+        }
 
-        crate::asm::dsb();
-        crate::asm::isb();
+        // NOTE(unsafe): Having &mut self means no ISR, for example, is going to
+        // try to alter this register at the same time.
```
adamgreig, line 415: The safety is provided by the assembly function entering a critical section, so it doesn't matter that we had a unique reference to `self` anyway. Perhaps this comment and the one below could be combined into a single comment that the asm routine contains a critical section so its access to SCB cannot be raced and it contains the required barriers for cache maintenance operation (and same for icache).

In fact, both methods could be made static now (albeit a breaking change) as neither require access to SCB any more.

cbiffle: I can merge the comments.

Making the methods static is a design decision not related to the implementation -- it could have been done before, since the asm code is doing the equivalent of `unsafe { &*SCB::ptr() }`. So it comes down to your preference: should any code anywhere be able to turn on caches as "ambient authority," without having been handed an enabling token? It feels inconsistent with the rest of the API, which tends to _really want_ you to have been passed a phantom peripheral representation to do things.

In either case, I'd prefer to get this fix available without bumping major versions, so making the operations static could wait?

cbiffle: (For example, the data cache invalidation/cleaning operations are necessary in cases where you have a CPU with caches, and another device, sharing control of a block of memory, e.g. for DMA, if you haven't simply disabled caching for that section of memory. Right now, the only code with access to these operations must have exclusive control over the entire `SCB`. Assuming this was deliberate, the cache enable operations are strictly less common than other maintenance operations, and should probably not be the first or only ones to become `static` if we want to `static` some things.)

adamgreig: I would say our current requirement to have the peripheral token exists in order to use Rust's lifetime system to ensure exclusive access to the peripheral in aid of preventing data races in read-modify-write operations, rather than as a means of safety-enforced access control to choose what code can make system modifications, if that makes sense. We have a number of static methods already and in general they're more ergonomic.

I completely take your point that clean and invalidate ops are the ones that really want to be static though (currently I have to unsafely create an SCB to use them since it's not practical to pass the only SCB instance into e.g. an Ethernet driver). Perhaps let's leave these operations as they are for now and consider making the majority of cache control static later.

Currently our master branch already contains breaking changes to our latest release, but I'm happy to issue a patch release on the current 0.6 series containing this and other non-breaking bug fixes.

Issue #236, opened 2020-07-05
`cortex_m::asm::delay` is wrong on anything more complex than an M4


cbiffle:
The contract for `delay` currently states:

> Blocks the program for *at least* `n` instruction cycles

The current implementation is:

```
            llvm_asm!("1:
                  nop
                  subs $0, $$1
                  bne.n 1b"
                 : "+r"(_n / 4 + 1)
                 :
                 : "cpsr"
                 : "volatile");
```

This is approximately an assembly gloss of the code

```rust
let mut i = n / 4 + 1;
loop {
    i -= 1;
    if i == 0 { break }
}
```

Why the divide-by-four? Well, the code appears to be assuming that the minimum amount of time that the instruction sequence could take is four cycles, and so it is taking that into account to minimize the amount of over-delaying. That cycle count is approximately accurate on M3/M4 (it's an underestimate in practice, even in zero-wait-state RAM), but is wrong on M7 or any other part with any combination of nop-folding, branch prediction, and multiple issue. As a result, on such CPUs, this function delays for substantially fewer cycles than advertised -- by roughly a factor of 2, depending on cache and memory system configuration, branch alignment, etc. (It also appears to be wrong on M33, but M33 is not superscalar, so it's probably just an instruction timing issue.)

As CPU core complexity grows, any hand-tuned cycle delay loops need to be specialized for the particular microarchitecture being used, much like in the 486/Pentium era. Alternatively, you could use a conservative implementation that, when asked to loop for at least `n` cycles, repeats the loop `n` times -- that should be safe _for now_, but I wouldn't recommend it, particularly because we've got the zero-overhead-loop extension to ARMv8-M on the horizon.

Shout out to @labbott for noticing this.

adamgreig:
Even beyond the microarchitecture or CPU model (e.g. `CPUID.PartNo`), if the current cache configuration, what memory the code is being read from, branch prediction, etc, all have a strong effect on cycles-per-loop it seems we can't hope to get a good bound here.

Perhaps the most correct solution is to update the documentation to indicate that the method delays for O(n) cycles, or allow user customisation of the scale factor?

therealprof:
Hm, 4 is indeed an odd choice. I'd probably left the `nop` out and gone for 2. I don't think you can execute the loop in less than 2 cycles on any current MCU, can you?

cbiffle:
[...]

PR #240, opened 2020-07-08
Better delay description

therealprof:
CC #236

### DIFF
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -30,7 +30,7 @@ pub fn bkpt() {
 /// level, however it is dependent on the specific architecture and core configuration.
 ///
 /// NOTE that the delay can take much longer if interrupts are serviced during its execution 
-/// and the execution time may wary with other factors. This delay is mainly useful for simple
+/// and the execution time may vary with other factors. This delay is mainly useful for simple
 /// timer-less initialisation of peripherals iff accurate timing is not essential. In any other
 /// case please use more accurate methods to cause a delay.
 #[inline]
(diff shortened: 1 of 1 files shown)


src/asm.rs
```diff
@@ -26,13 +26,16 @@ pub fn bkpt() {
...
 /// Blocks the program for *at least* `n` instruction cycles
 ///
-/// This is implemented in assembly so its execution time is the same regardless of the optimization
-/// level.
+/// This is implemented in assembly so its execution time is independent of the optimization
+/// level, however it is dependent on the specific architecture and core configuration.
 ///
-/// NOTE that the delay can take much longer if interrupts are serviced during its execution.
+/// NOTE that the delay can take much longer if interrupts are serviced during its execution 
+/// and the execution time may wary with other factors. This delay is mainly useful for simple
```
jonas-schievink, line 33:
```suggestion
/// and the execution time may vary with other factors. This delay is mainly useful for simple
```

src/asm.rs
```diff
@@ -26,13 +26,16 @@ pub fn bkpt() {
...
+/// timer-less initialisation of peripherals iff accurate timing is not essential. In any other
```
jonas-schievink, line 34:
```suggestion
/// timer-less initialization of peripherals if accurate timing is not essential. In any other
```

therealprof: Hm, the **iff** is intended. I'm fine with changing the spelling to AE.

jonas-schievink: Maybe expand it to "if and only if" then? *iff* tends to confuse people.

therealprof: Hm, I've never met anyone who was confused about that but sure, why not.

PR #241, opened 2020-07-09
Add new InterruptNumber trait

adamgreig:
This is a first go at the new trait needed for https://github.com/rust-embedded/svd2rust/pull/455 since we removed `Nr` from bare-metal.

In this case I've written it as `unsafe trait InterruptNumber: Into<u16>` rather than providing a conversion method inside the trait; I think this is neat and idiomatic but please correct me if there's a reason to not do it like this.

[Here's](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4f2f8b9604b5a62298f9907780d844c7) a playground link showing an example implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- New `InterruptNumber` trait is now required on interrupt arguments to the
+  various NVIC functions, replacing the previous use of `Nr` from bare-metal.
+
 ## [v0.6.2] - 2020-01-12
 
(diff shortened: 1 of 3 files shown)


therealprof:
I like it! I'm also a big fan of the `Into<u16>` idea. 😉

adamgreig:
I've updated to fix the failing CI (due to the use of some `&InterruptNumber` types; Into<u16> isn't implemented on reference types). The new trait now requires `Copy` and I've removed use of references.

therealprof:
As discussed in todays meeting this should be adjusted to lose the `Into<u16>` trick and provide a `number` method instead. Everything else seems ready to go.

Issue #242, opened 2020-07-12
Remove `#[no_mangle]` from `CORE_PERIPHERALS` to enable zero-cost peripherals


jonas-schievink:
The `#[no_mangle]` on the `CORE_PERIPHERALS` currently prevents LLVM from optimizing it out when only stored to:

https://github.com/rust-embedded/cortex-m/blob/1c9d36ebe682d79f35ef2b1f34e611cfbf8ee81e/src/peripheral/mod.rs#L152-L153

This means that even if an app only uses `Peripherals::steal()`, it still has to pay 1 Byte of RAM for the global. Since 1 is greater than 0, this can be improved.

`#[no_mangle]` is used to prevent multiple cortex-m versions to be linked together, so this behavior must be retained somehow. While we already use `links` in `Cargo.toml` for this, this only works with newer versions (that are currently not even released), so another (zero-sized) global should be introduced to take over `CORE_PERIPHERALS`' job.

Issue #239, opened 2020-07-06
Proposal for separating cortex-m into smaller parts


cbiffle:
# Summary

The `cortex-m` crate provides facilities for using peripherals and special
instructions on the ARM Cortex-M series. This makes it useful in a wide range of
embedded applications. However, the crate also does a few other things that
restrict its applicability. I would like to propose finding a seam in the crate,
between its low-level generally-applicable bits and its higher-level bits, and
separating it into two.

# Context

Most Rust applications using the `cortex-m` crate are what I would call
**monolithic** applications: they consist of a single program, compiled/linked
together, and do not use memory protection to isolate components, nor the
processor's privileged/unprivileged distinction to isolate a kernel.

There are (for the sake of this discussion) three _other_ kinds of Cortex-M
applications, however: those that use limited memory protection and privilege
within a single linked application (FreeRTOS `CM3_MPU` port); those that isolate
a kernel and run drivers in privileged mode (Tock; uCLinux and programs running
atop it); and those that use memory protection to run drivers in unprivileged
mode (so-called "multiserver" systems). The current APIs aren't well suited for
any of these applications.

As an example, consider issue #223. The memory safety properties of the current
`cortex-m` API rely on critical sections, and they turn out to be void in
unprivileged mode -- because the library assumes a particular method of
implementing critical sections.

# A taxonomy of `cortex-m` APIs

I've been thinking about this a lot recently, while using `cortex-m` in both the
kernel and userland of an isolated multiserver operating system (not yet
released). I see the following broad groups of functionality. Each of these is
probably not a separate crate, to be clear, these are merely categories.

## The Universally Relevant

Things that make sense to use in a kernel, in userland, with or without
protection, etc. Stuff in this category should include:

1. Operations that are reasonable in both privileged and unprivileged mode.
   (Note: operations that reliably trap in unprivileged mode are OK. MSR, MRS,
   and CPS do _not_ trap and should be exposed only very carefully.)

2. Operations that do not assume a single linked program (e.g. they must not use
   a `static` to coordinate access to a shared resource).

This list may be really short; even with MPU shenanigans, you cannot make the
peripherals on the Private Peripheral Bus (which is to say, most of those
defined in `cortex-m`) accessible to unprivileged code.

At first glance, here are some things that belong in this level:

- `wfi`/`wfe`, probably `sev`
- `bkpt`
- `delay`, were it correct (see #236)
- `nop`
- `udf`
- `isb`/`dsb`/`dmb`
[...]

PR #243, opened 2020-07-12
Allow the taken flag to be optimized out

jonas-schievink:
Fixes https://github.com/rust-embedded/cortex-m/issues/242

Verified manually that it still prevents linking in multiple versions.


therealprof:
> However I do note that #239 still has the same issue that both an operating system and an application/driver may need to use cortex-m, but be linked totally separately and only combined at runtime, in which case this "safe" interface is not; perhaps ultimately cortex-m should lose `take()` and the singleton safety be provided by cortex-m-rt or other runtime.

I agree.

PR #246, opened 2020-07-16
Always link pre-built asm, required for new cache management functions

### LINKED ISSUE #245: enable_icache gives linker error
Hey all,

when enabling the ICache, I now get this linker error:
image

It seems it's got to do with the change in #234.
The __enable_icache symbol that's defined in the assembly can't be found.

I'm compiling for the stm32h743 on the latest nightly (rustc 1.46.0-nightly (346aec9b0 2020-07-11)).

My .cargo/config (it's just the default I believe):
```
[target.thumbv7em-none-eabihf]
runner = 'arm-none-eabi-gdb -x debug.gdb'
rustflags = [
  # LLD (shipped with the Rust toolchain) is used as the default linker
  "-C", "link-arg=-Tlink.x",

  # if you run into problems with LLD switch to the GNU linker by commenting out
  # this line
  #"-C", "linker=arm-none-eabi-ld",

  # if you need to link to pre-compiled C libraries provided by a C toolchain
  # use GCC as the linker by commenting out both lines above and then
  # uncommenting the three lines below
  # "-C", "linker=arm-none-eabi-gcc",
  # "-C", "link-arg=-Wl,-Tlink.x",
  # "-C", "link-arg=-nostartfiles",
]

[build]
target = "thumbv7em-none-eabihf"
```
I've tried the commented options as well, but that didn't help.

adamgreig:
In #234 we added new `__enable_icache` and `__enable_dcache` assembly routines, but without providing an inline assembly version, to reduce duplication and since there wasn't an expected performance impact. However, our build.rs currently only links the pre-built object if the `inline-asm` feature is disabled, which means currently you can't call `enable_icache()` and use `inline-asm` at the same time.

This PR makes us always link against the pre-built objects (for thumb targets) even if `inline-asm` is used; the pre-built object would only be used for the cache management routines at present but we may want to put more routines into the assembly blob only in the future.

Closes #245.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,7 +6,7 @@ fn main() {
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     let name = env::var("CARGO_PKG_NAME").unwrap();
 
-    if target.starts_with("thumb") && env::var_os("CARGO_FEATURE_INLINE_ASM").is_none() {
+    if target.starts_with("thumb") {
         fs::copy(
             format!("bin/{}.a", target),
             out_dir.join(format!("lib{}.a", name)),


jonas-schievink:
What was the original motivation for not linking them when `inline-asm` is turned off? Are there any impacts to binary size or something?

diondokter:
Compiled based on the commit and it works for me.

adamgreig:
> What was the original motivation for not linking them when `inline-asm` is turned off? Are there any impacts to binary size or something?

Everything in them was previously implemented with inline asm, so there wasn't any need to have them. LTO seems to remove the unused functions fine.

PR #248, opened 2020-07-19
v0.6.3 backports

jonas-schievink:
Doesn't include *everything* we could, but I ran out of steam. The important stuff should be there.

Next steps:
- changelog
- bump version and release

PR #249, opened 2020-07-20
More v0.6.3 backports

adamgreig:
I think this includes the remainder of the non-breaking changes since v0.6.2, with a couple of exceptions:

* https://github.com/rust-embedded/cortex-m/pull/240 seemed low-impact but had loads of separate commits to cherry-pick
* https://github.com/rust-embedded/cortex-m/pull/220 is I think non-breaking but was quite a substantial change, perhaps I could still include it

~~I did include https://github.com/rust-embedded/cortex-m/pull/226 which adds a new field to `Peripherals`, but as I understand it that should be a non-breaking change since it's non-exhaustive.~~

I've updated the CHANGELOG with all the changes from this PR and the previous #248.


jonas-schievink:
> I did include #226 which adds a new field to `Peripherals`, but as I understand it that should be a non-breaking change since it's non-exhaustive.

I think it's only non-exhaustive on master, not 0.6.x (if not, did I accidentally backport that breaking change?)

adamgreig:
Aha, no, I merged the clippy fix (`manual_non_exhaustive`) but it's still non-exhaustive on 0.6.x. I'll remove #226.

adamgreig:
Looks like the -Dwarnings in CI is causing all nightlies to fail due to a clippy lint? [here](https://travis-ci.org/github/rust-embedded/cortex-m/jobs/709955783#L365).

therealprof:
> Looks like the -Dwarnings in CI is causing all nightlies to fail due to a clippy lint? [here](https://travis-ci.org/github/rust-embedded/cortex-m/jobs/709955783#L365).

I wonder why it's not failing on stable, too.

adamgreig:
The lint doesn't seem quite right, either, I don't think it's understood what we're doing. I guess let's not worry about it for 0.6.x and solve it in master which presumably has the same issue?

therealprof:
It can't see through the `#[cfg]` gate.

We have that in a few HAL impls, too: If you have a `match` and all arms except for 1 or 2 are cfg-gated it'll recommend an `if` instead. Unfortunately you cannot `#[cfg]` gate an `if` expression or we'd have switched over to that instead...

PR #253, opened 2020-08-01
Clean up some unsafe code introduced by #202

bugadani:
Although I haven't checked these exact functions, a simple clone seems to be enough to elide the bounds checks: https://godbolt.org/z/a1jb4h (`access` taken directly from [the issue](https://github.com/rust-lang/rust/issues/13926)).


bugadani:
Actually, deriving Copy along with clone messes things up: https://godbolt.org/z/zhYfEs

Issue #254, opened 2020-08-17
check-blobs.sh does nothing


jonas-schievink:
Running it modifies all the archives on my system, but doesn't fail.

Since we've started passing `-g` to the assembler, all files now contain system-specific paths and are not reproducible.

adamgreig:
`check-blobs.sh` compares the output of `arm-none-eabi-objdump -Cd`, which does not contain the system-specific paths and the CI seems to be correctly checking that those outputs match, but unfortunately that's not the same as checking the actual `.a` files match byte-for-byte. Can we specify `-g` without causing specific paths to be included?

PR #255, opened 2020-08-18
Remap debuginfo paths

jonas-schievink:
cc https://github.com/rust-embedded/cortex-m/issues/254


therealprof (approved):
Whee, that looks gnarly. Looks good on a first glance but the result could use a second look.

bors r+

therealprof:
I had a peek into the generated binaries and looks fine to me.

Issue #217, opened 2020-05-19
v0.6.3 release?


japaric:
Did anything breaking land in master since the v0.6.2 release?

I would like to see a v0.6.x release that includes binary-level fixes like #212 and #216

Disasm:
There was a breaking change since v0.6.2: https://github.com/rust-embedded/cortex-m/pull/190

little-arhat:
0.6.2 no longer works in nightly (`rustc 1.46.0-nightly (118b50524 2020-06-06)`) due to `asm!` deprecation (was fixed in https://github.com/rust-embedded/cortex-m/issues/204).

jonas-schievink:
It should still work without the unstable `inline-asm` feature

little-arhat:
@jonas-schievink 

```
When this feature is enabled the implementation of all the functions inside the asm and register modules use inline assembly (asm!) instead of external assembly (FFI into separate assembly files pre-compiled using arm-none-eabi-gcc). The advantages of enabling inline-asm are:

    Reduced overhead. FFI eliminates the possibility of inlining so all operations include a function call overhead when inline-asm is not enabled.

    Some of the register API only becomes available only when inline-asm is enabled. Check the API docs for details.

The disadvantage is that inline-asm requires a nightly toolchain.
```

Documentation doesn't say this feature is unstable.

adamgreig:
I think we should start looking in to a 0.6.3 release with only the non-breaking changes and probably a 0.7.0 series with the breaking changes; as far as I'm aware the breaking PRs since 0.6.2 are #180 #181 #190 #191 #192 #193.

kanishkarj:
It would be really great if `0.6.3` comes out soon. I want to actually publish a crate but the docs don't build because it uses the latest nightly. And it is not possible on the latest nightly like @little-arhat pointed out.

jonas-schievink:
I'll take this on

adamgreig:
0.6.3 is now published :tada:

kanishkarj:
Thanks a lot!!

hug-dev:
Thanks for the release!

A small question: the `cmse` module does not seem to appear in the [docs](https://docs.rs/cortex-m/0.6.3/cortex_m/). Is that because it is behind the `armv8m` feature?

adamgreig:
[...]

PR #256, opened 2020-08-23
Add embedded targets to docs.rs

hug-dev:
Some modules of this repo are gated by the various targets and hence docs.rs does not show the documentation for them! Like the Armv8-M ones for [example](https://github.com/rust-embedded/cortex-m/issues/217#issuecomment-661039292) and the `cmse` module.

docs.rs now allows building for specific target and since [this PR](https://github.com/rust-lang/docs.rs/pull/633) even for targets available via `rustup`!

The `stm32f3_discovery` crate does it, look at ["Platform" on docs.rs](https://docs.rs/stm32f3-discovery/0.4.0/stm32f3_discovery/index.html).

So I think it would be very neat to have on `cortex-m` and I proposing to add the following targets. The first one will show by default, I choosed the latest and greatest for that 😄 

I tested locally with `cargo doc --target ...` for all of them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,17 @@ bare-metal = { version = "0.2.0", features = ["const-fn"] }
 volatile-register = "0.2.0"
 bitfield = "0.13.2"
 
+[package.metadata.docs.rs]
+targets = [
+    "thumbv8m.main-none-eabihf",
+    "thumbv6m-none-eabi",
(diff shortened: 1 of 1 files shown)


adamgreig:
> Could a quick patch release be done for that to be up on `docs.rs`  ? Or if there is a release that is going to be done in the next few weeks that's fine as well.

Hopefully we can get 0.7 out soon which will include this.

PR #257, opened 2020-08-23
Provide a delay driver based on SysTick 

Disasm:


src/delay.rs
```diff
@@ -0,0 +1,119 @@
+//! A delay driver based on SysTick
```
jonas-schievink, line 1:
```suggestion
//! A delay driver based on SysTick.
```

...etc for the other comments

thalesfragoso:
Should we add a prelude ?

src/prelude.rs
```diff
@@ -0,0 +1,3 @@
+//! Prelude
+
+pub use embedded_hal::prelude::*;
```
thalesfragoso, line 3: Should we rename the traits to something unique to avoid collisions with other `embedded-hal` exports ? Would that work okay ? In that case we would only re-export the delays traits.

Disasm: I checked this and it works even with collisions (double-imported from `stm32l4xx-hal` and `cortex-m`). I'm not sure it works like this on all rust versions >=MSRV, thugh.

Disasm: I checked on 1.31.0 with a fake hal and it works too. However, importing prelude does *not* work on 1.31.0 if `cortex-m` and hal use different versions of `embedded-hal` (I used `1.0.0-alpha.1` in my fake hal to check). Works fine on stable, though.

thalesfragoso: This seems fine them.

Issue #194, opened 2020-02-05
Remove Arm server access from CI


jonas-schievink:
We currently download the GCC toolchain from the Arm servers during CI, which is unreliable. We should try installing a toolchain via apt-get, since Travis mirrors that internally. Hopefully there's a version that works for us.

hug-dev:
You have [this PPA](https://launchpad.net/~team-gcc-arm-embedded/+archive/ubuntu/ppa) that might work. The version is a bit old but supports Armv8-M. It's the one used in the Rust repo `dist-various-1` [Dockerfile](https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-various-1/Dockerfile).

Issue #139, opened 2019-04-11
Support Linker Plugin to inline ASM


jamesmunns:
See https://github.com/Disasm/inline-test and https://doc.rust-lang.org/nightly/rustc/linker-plugin-lto.html

This should be stable shortly in Rust 1.34.0

CC @Disasm and @pftbest

jamesmunns:
Also CC @rust-embedded/cortex-m

jamesmunns:
As a note, I think this would entail the following steps:

* [ ] Move the contents of asm*.s into files called asm*.c, marking each C function as `__inline__(always)`
* [ ] Wrap the current assembly into explicit C functions, removing the return calls (e.g. `bx lr`), and respecting the C calling convention for functions that take arguments, such as `__psp_w`'s use of `r0`
* [ ] Modify the `assemble.sh` function to match [inline-test](https://github.com/Disasm/inline-test/blob/master/assemble.sh)'s behavior using clang to build with `-flto=thin`.
* [ ] Add documentation to instruct users to add `"-C", "linker-plugin-lto"` to their `.cargo/config`'s `rustflags` array.

As far as I know, these changes should be backwards compatible, and should (probably?) not cause regressions by switching from ASM to ASM-in-C. This is probably worth verifying. If it does NOT, I would suggest that we make this a non-breaking change. If it DOES, I would suggest that we add this as a new feature that can be opted-in to, typically only used when the user has added support for linker-plugin-lto

jamesmunns:
Follow up: I don't think this could be done without a new feature/breaking change. The pre-generated static archives would include LLVM bitcode, which would likely work with any version of rust based on LLVM7 or LLVM8 (I'm not sure if this includes Rust 1.32, which I think is our current MSRV), but this would almost certainly break any projects still using `arm-none-eabi-ld`.

I'd suggest that we make this an opt-in feature, and ship both sets of static archives for now.

Disasm:
Good news: binary archives compiled with bitcode object files work "as is" in our usual build scenario (without ThinLTO and even LTO).
Bad news: this works only on the specific range of Rust versions compatible with this bitcode.
Another issue is that you never know the last Rust version with the compatible LLVM built-in, so you can't check for the upper supported Rust version in build.rs.

I suggest adding an unstable `lto` feature that will enable these LTO-ready binary archives instead of the standard ones. This way a user could enable ThinLTO for a specific environment.

PR #259, opened 2020-08-25
Rust assembly stubs

jonas-schievink:
Disclaimer: I have not tested any of this (help wanted, we have no tests in here).

Thanks to @therealprof for nerd-sniping me into oblivion.

Fixes https://github.com/rust-embedded/cortex-m/issues/254
Fixes https://github.com/rust-embedded/cortex-m/issues/194
Fixes https://github.com/rust-embedded/cortex-m/issues/139

# Summary

* Remove the assembly files in favor of a new `asm.rs`, which uses unstable inline assembly and provides a C ABI interface.
* Replace the shell scripts by a [`cargo-xtask`](https://github.com/matklad/cargo-xtask/).
* While we're at it, also pre-build artifacts that are compatible with linker-plugin LTO, fixing https://github.com/rust-embedded/cortex-m/issues/139 (again, not tested)

This means that contributors and maintainers just need ~~a nightly Rust compiler installed~~ to run `cargo xtask assemble`. No binutils, no assembler, no `ar`, no GCC/Clang, and especially nothing from the godawful Arm servers, fixing https://github.com/rust-embedded/cortex-m/issues/194. You don't even have to install the correct nightly Rust toolchain, `cargo xtask` does it for you (and installs all the thumb targets too).


ci/script.sh
```diff
@@ -29,7 +29,7 @@ main() {
     esac
 
     if [ $TARGET = x86_64-unknown-linux-gnu ]; then
-        ./check-blobs.sh
+        cargo xtask check-blobs
```
jonas-schievink, line 32: Hmm, this might break between nightly updates when the version changes. That probably means we'll have to pin the nightly version to use for the blobs, which is probably a good idea anyways.

therealprof: Yes, I agree, we should definitely pin a version so we don't risk all kinds of funky random issues.

thalesfragoso:
Great work, I did some tests here, targeting the `plugin-lto`. First, it seems that if I use `-Clinker-plugin-lto` I can't compile with `opt-level= "s"` (and probably "z"), because I get this error:

```console
rust-lld: error: -plugin-opt=Os: number expected, but got 's'
```

The following tests used `opt-level=3`, I wanted to try our `Mutex` with `plugin-lto`, so here is the code:

```rust
static FLAG: Mutex<Cell<bool>> = Mutex::new(Cell::new(false));

#[entry]
fn main() -> ! {
    while !free(|cs| FLAG.borrow(cs).get()) {}

    unsafe {
        ptr::write_volatile(0x40000000 as *mut u32, 1);
    }
    loop {
        asm::nop();
    }
}

#[exception]
fn SysTick() {
    free(|cs| FLAG.borrow(cs).set(true));
}
```

With `cortex-m` from crates.io:

```asm
Dump of assembler code for function main:
   0x08000400 <+0>:     push    {r7, lr}
   0x08000402 <+2>:     mov     r7, sp
   0x08000404 <+4>:     bl      0x800040a <plugin_test::__cortex_m_rt_main>
   0x08000408 <+8>:     udf     #254    ; 0xfe
End of assembler dump.

Dump of assembler code for function plugin_test::__cortex_m_rt_main:
   0x0800040a <+0>:     push    {r7, lr}
   0x0800040c <+2>:     mov     r7, sp
   0x0800040e <+4>:     movw    r5, #0
   0x08000412 <+8>:     movt    r5, #8192       ; 0x2000
   0x08000416 <+12>:    bl      0x800050c <__primask>
   0x0800041a <+16>:    mov     r4, r0
   0x0800041c <+18>:    bl      0x8000502 <__cpsid>
   0x08000420 <+22>:    lsls    r0, r4, #31
   0x08000422 <+24>:    ldrb    r6, [r5, #0]
   0x08000424 <+26>:    it      eq
   0x08000426 <+28>:    bleq    0x8000506 <__cpsie>
   0x0800042a <+32>:    cmp     r6, #0
   0x0800042c <+34>:    beq.n   0x8000416 <plugin_test::__cortex_m_rt_main+12>
   0x0800042e <+36>:    mov.w   r0, #1073741824 ; 0x40000000
   0x08000432 <+40>:    movs    r1, #1
   0x08000434 <+42>:    str     r1, [r0, #0]
   0x08000436 <+44>:    bl      0x800050a <__nop>
   0x0800043a <+48>:    b.n     0x8000436 <plugin_test::__cortex_m_rt_main+44>
End of assembler dump.
```

With this branch without plugin-lto:

```asm
Dump of assembler code for function main:
   0x08000400 <+0>:     push    {r7, lr}
   0x08000402 <+2>:     mov     r7, sp
   0x08000404 <+4>:     bl      0x800040a <plugin_test::__cortex_m_rt_main>
   0x08000408 <+8>:     udf     #254    ; 0xfe
End of assembler dump.

Dump of assembler code for function plugin_test::__cortex_m_rt_main:
   0x0800040a <+0>:     push    {r7, lr}
   0x0800040c <+2>:     mov     r7, sp
   0x0800040e <+4>:     movw    r5, #0
   0x08000412 <+8>:     movt    r5, #8192       ; 0x2000
   0x08000416 <+12>:    bl      0x800050e <asm::__primask>
   0x0800041a <+16>:    mov     r4, r0
   0x0800041c <+18>:    bl      0x8000502 <asm::__cpsid>
   0x08000420 <+22>:    lsls    r0, r4, #31
   0x08000422 <+24>:    ldrb    r6, [r5, #0]
   0x08000424 <+26>:    it      eq
   0x08000426 <+28>:    bleq    0x8000506 <asm::__cpsie>
   0x0800042a <+32>:    cmp     r6, #0
   0x0800042c <+34>:    beq.n   0x8000416 <plugin_test::__cortex_m_rt_main+12>
   0x0800042e <+36>:    mov.w   r0, #1073741824 ; 0x40000000
   0x08000432 <+40>:    movs    r1, #1
   0x08000434 <+42>:    str     r1, [r0, #0]
   0x08000436 <+44>:    bl      0x800050a <asm::__nop>
   0x0800043a <+48>:    b.n     0x8000436 <plugin_test::__cortex_m_rt_main+44>
End of assembler dump.
```

Pretty much the same thing, only the address of `primask()` that is a bit different.

Now with `plugin-lto`:

```asm
Dump of assembler code for function main:
   0x08000400 <+0>:     push    {r7, lr}
   0x08000402 <+2>:     mov     r7, sp
   0x08000404 <+4>:     bl      0x8000408 <plugin_test::__cortex_m_rt_main>
End of assembler dump.

Dump of assembler code for function plugin_test::__cortex_m_rt_main:
   0x08000408 <+0>:     push    {r7, lr}
   0x0800040a <+2>:     mov     r7, sp
   0x0800040c <+4>:     movw    r0, #0
   0x08000410 <+8>:     movt    r0, #8192       ; 0x2000
   0x08000414 <+12>:    b.n     0x8000418 <plugin_test::__cortex_m_rt_main+16>
   0x08000416 <+14>:    cbnz    r1, 0x8000428 <plugin_test::__cortex_m_rt_main+32>
   0x08000418 <+16>:    mrs     r2, PRIMASK
   0x0800041c <+20>:    cpsid   i
   0x0800041e <+22>:    ldrb    r1, [r0, #0]
   0x08000420 <+24>:    lsls    r2, r2, #31
   0x08000422 <+26>:    bne.n   0x8000416 <plugin_test::__cortex_m_rt_main+14>
   0x08000424 <+28>:    cpsie   i
   0x08000426 <+30>:    b.n     0x8000416 <plugin_test::__cortex_m_rt_main+14>
   0x08000428 <+32>:    mov.w   r0, #1073741824 ; 0x40000000
   0x0800042c <+36>:    movs    r1, #1
   0x0800042e <+38>:    str     r1, [r0, #0]
   0x08000430 <+40>:    nop
   0x08000432 <+42>:    b.n     0x8000430 <plugin_test::__cortex_m_rt_main+40>
End of assembler dump.
```

So everything seems to be working great, at least in this example. The `plugin-lto` managed to do the inlining and some more optimizations because of that, but it didn't optimize our read to the `Mutex`. I expect that we can rely on that, because the `asm!()` macro does clobbering.

ci/script.sh
jonas-schievink: Added an `asm-toolchain` file to pin the nightly version. `cargo xtask` will install it automatically.

Cargo.toml
```diff
@@ -21,6 +21,19 @@ volatile-register = "0.2.0"
...
 
+[features]
+cm7-r0p1 = []
+inline-asm = []
+
+# Link against prebuilt assembly blobs that are compatible with Linker-Plugin LTO.
+# When using this feature, you *must* pass `-Clinker-plugin-lto` to rustc. See this link for
+# details:
+# https://doc.rust-lang.org/stable/rustc/linker-plugin-lto.html
+linker-plugin-lto = []
```
thalesfragoso, line 32: Could you also add that it isn't possible to use this feature on opt-level s/z ?

Disasm:
I suspect that linking to `-lto` archives will stop working after then next LLVM version bump in Rust (and probably with old Rust versions too), but otherwise this approach looks great :)

jonas-schievink:
From what I gathered, ThinLTO summaries are at least forwards-compatible, so we should be able to keep using old archives with newer Rust versions. I haven't tested this though (not sure if there's any way to do that).

Disasm:
@jonas-schievink One problem I noticed while experimenting with LLVM binaries for ThinLTO is that sometimes even data layout string changes, making the binary incompatible with other LLVM versions. Probably it's a good idea to have multiple binaries, one for each LLVM version. We can even fallback to regular .o files when there is no `-lto` archive for the "current" LLVM version.

jonas-schievink:
I don't really want to set that up, if users really want their asm inlined they can use a Rust version that's compatible

therealprof:
Yeah, I'd rather mark LTO as an experimental feature for now and not block this very nice improvement.

Cargo.toml
jonas-schievink: Moved the docs into lib.rs and expanded them

therealprof:
Looks good to me, I'd have made the disclaimer a bit more explicit but hey.

Anything else have any opinion on this? Otherwise I'd be tempted to approve it.

PR #261, opened 2020-08-28
Replace shell scripts with integration test

jonas-schievink:
* Removes the Clippy check from CI (I don't think we should fail CI when Clippy finds something, it seems better to use the GitHub Actions plugin)
* Tests all targets (thumbv8m.base was missing)
* Bumps MSRV to 1.35.0 (to get the thumbv8m.base libcore)


jonas-schievink:
Hmm, do we not configure bors to look for the travis builders?

therealprof (approved):
LGTM

bors r+

PR #71, opened 2017-12-23
map asm! ops to unimplemented! on non ARM targets

japaric:
closes #63
cc @hannobraun

PR #262, opened 2020-08-29
Merge asm implementations

jonas-schievink:
This replaces the implementation of `inline-asm` with the file I wrote in #259 (and some fixes).

All functions that call assembly now do so via a `call_asm!` macro that either dispatches to a call to an `#[inline(always)]` function containing the inline `asm!`, or to the FFI shim. This makes all functions that call into asm significantly shorter.

The FFI shim is now also macro-generated, which makes it very small.


adamgreig (approved):
This looks great, nice work! I'll hold off on the merge to give everyone a chance to have a look over it. Once it's merged let's sort out the remaining fixmes/todos and I'd like to modify the methods that use explicit registers to instead use generic named registers that the compiler can allocate.

therealprof (approved):
Nice!

asm/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn __primask_r() -> u32;
+    fn __psp_r() -> u32;
+    fn __psp_w(val: u32);
+    fn __sev();
+    fn __udf();
+    fn __wfe();
+    fn __wfi();
+}
+
+#[cfg(armv7m)]
```
jonas-schievink, line 76:
```suggestion
// v7m *AND* v8m.main, but *NOT* v8m.base 
#[cfg(any(armv7m, armv8m_main))]
```

src/asm.rs
```diff
@@ -1,179 +1,66 @@
...
-            extern "C" {
-                fn __bkpt();
-            }
-
-            __bkpt();
-        },
-
-        #[cfg(not(cortex_m))]
-        () => unimplemented!(),
-    }
```
Rahix, line 24: I recently dug this up for unrelated reasons.  I just wanted to mention, this match construct was introduced because of rust-embedded/cortex-m#63 with the actual change being from rust-embedded/cortex-m#71.  Is it no longer relevant?

jonas-schievink: It still builds fine for me. Maybe not with `inline-asm`. I should be able to add this feature back though.

jonas-schievink: Ah, I remember now, I removed this to avoid all the warnings this generates. These were the reason almost the entire crate used function arguments like `_val`. I think we should leave this as-is for now.

therealprof (approved):

therealprof (changes requested):
Actually it looks like the MSRV has changed again, yet README.md still says 1.31

therealprof (approved):
Thanks!

bors r+

PR #263, opened 2020-09-01
Import cortex-m-semihosting and panic-semihosting into this repo

jonas-schievink:
Motivation:

* Allows writing QEMU tests for `cortex-m`'s functionality that use semihosting to control QEMU. Previously these crates would pull in another cortex-m version, which doesn't work. Now they have a `path` dependency on the root crate.
* Lets us share the outline-inline-assembly setup and `cargo-xtask` in general.
* Lets us share CI and bot setup between more crates.
* 2 fewer repos to triage and keep track of (I'll transfer their issues after this is merged).

I also want to import cortex-m-rt, but I'll do that in a later PR.

CI was updated to build-test all crates with all or most feature combinations, like it did before.


therealprof:
I am still very much not a fan of this and would like to put this up for a vote.

adamgreig:
PR looks fine content-wise. @therealprof, I know you weren't keen on having semihosting be part of cortex-m itself, but what's the issue with having the crates live inside this repo? We save on management and overhead, but it doesn't change anything for users. Is the objection just to having any semihosting support at all?

therealprof:
I think it doesn't feel right from an architectural standpoint to depend on the semihosting crate and a panic handler just for CI purposes.

Moving them into this repository blesses the semihosting approach and highlights it in a weird way.

I'd rather we only move the `syscall` into `cortex-m` and use it directly.

adamgreig:
How else would you test inside qemu?

therealprof:
> How else would you test inside qemu?

Not even sure what we want to test in `cortex-m` that has no other choice than to use semihosting. Even this PR doesn't seem to actually add any semihosting dependency for testing purposes (despite saying so in the motivation) so it's hard to tell how to eliminate such a dependency.

jonas-schievink:
Semihosting is, to my knowledge, the only way to terminate QEMU with a success error code, and the only *clean* way to terminate it with an error code (the non-clean way is to cause a lockup). It is also the simplest way to print to stdout, which is useful for doing snapshot testing outside of QEMU (I'm not aware of any others that work on all targets we support).

I did not add any tests to this PR because I wanted to land the basic prerequisites first (and tests are quite distinct from what this PR aims to do), but if even that ends up being so controversial I'm not sure if my time isn't better spent elsewhere.

> Moving them into this repository blesses the semihosting approach and highlights it in a weird way.

I think it is now highlighted much less than before, when it had its own dedicated repo and issue tracker.

thalesfragoso:
I'm in favor of this change. I don't think this is putting semihosting at a spotlight or something, the people looking at our CI probably already know its disadvantages and alternatives.

If we wish to discourage beginners to use semihosting, our time would be better spent removing it from examples, which is arguably already happening and we have a lot of other tools in the spotlight now.

Also, I don't understand why we should go out of our way to look for possible alternatives when the work is already done and it matches our requirements.

Anyway, this is my opinion on the matter, thanks for your work Jonas.

therealprof:
> > I don't see why we couldn't just use the syscall functionality to that effect.

> That's what this _does_, essentially; it's providing the wrapper code around `syscall` that allows it to be used to communicate with qemu (or a debug probe).

The wrapper and panic handler go way beyond providing lowlevel primitives which is what this repo should focus on IMHO. The only motiviation mentioned to break out of this idea is to enable CI which this PR doesn't even do.

At the very least I'd like to see how this is going to be used to maybe come up with a possible alternative.

> Semihosting's an official ARM platform standard, so I think it makes perfect sense to support it inside cortex-m.

For me `cortex-m` is lowest level primitives and should remain at that level, this PR provides a higher level abstraction. There're tons of other ARM platform standards too, if we add higher level semihosting support those should have a rigt to be added as well, the very first thing coming to mind here are other panic-handler implementations (e.g. `panic-itm`), so `panic-semihosting` doesn't create the impression of being a blessed cocept.

> What issues do you think this will cause compared to having them separate, either for us or for users?

I do think it makes it harder for us to discourage the use of semihosting if we cannot keep the distance ourselves and accept bug reports for it in our central platform support repository.
 
> but organisationally this move seems to save us work and effort and remove duplication of CI/test infra

While this might save us an unquantified bit of duplicate work it also increases the risk of CI breakage due to some semihosting related breakage. All of this stuff is very low-level and somewhat fragile to compiler changes and whatnot; increasing the complexity is not going to help here.

adamgreig:
> The wrapper and panic handler go way beyond providing lowlevel primitives which is what this repo should focus on IMHO. The only motiviation mentioned to break out of this idea is to enable CI which this PR doesn't even do.

We already provide `iprintln` and other ITM-related functionality here; `hprintln` and related machinery are at the same level. Once these changes land, it becomes possible for the first time to add some CI tests which actually execute cortex-m code on an emulated thumb platform.

> the very first thing coming to mind here are other panic-handler implementations (e.g. panic-itm)

Especially given as ITM stuff is already in cortex-m, I'd be entirely in favour of putting panic-itm inside cortex-m. It makes it even easier for cortex-m users to specify a provided panic handler. We could even put panic-abort inside and thus reduce the dependency count of many apps one further. I see `cortex-m` not as the "lowest possible level" of abstractions - that would be a sort of "cortex-m-pac" which doesn't exist - but as a comprehensive library of functionality for the platform. We already provide a lot of high level utilities, and I don't see the advantage to splitting the others out into separate crates, either to us as maintainers or to users.

> While this might save us an unquantified bit of duplicate work it also increases the risk of CI breakage due to some semihosting related breakage. All of this stuff is very low-level and somewhat fragile to compiler changes and whatnot; increasing the complexity is not going to help here.

We already support cortex-m-semihosting as an official working group crate and repository, so any such breakage would just happen there instead and still need to be dealt with. Importing it here seems like a net win.

therealprof:
Sorry, completely forgot about this still being open.

I could find solace by moving a variety panic handlers into `cortex-m`. However are they all Cortex-M only?

There would be a few more points I could bring to the table re maintenance and (undesirably) bringing a spotlight to semihosting but if we integrate all sorts of panic handlers I could live with the maintenance complications and my main concern would be resolved.

adamgreig:
> I could find solace by moving a variety panic handlers into cortex-m. However are they all Cortex-M only?

They're all so simple that honestly just having a copy in cortex-m doesn't seem like a problem. It provides a useful feature to users of the crate and makes quickly adopting some sort of default panic handler easier.

I'd propose in addition to semihosting we include itm (using existing iprintln), abort/halt (infinite loop), and perhaps fault (jump to hardfault). Could those be a new PR so we can get this one moving?

therealprof:
> I'd propose in addition to semihosting we include itm (using existing iprintln), abort/halt (infinite loop), and perhaps fault (jump to hardfault). Could those be a new PR so we can get this one moving?

Fine. Who's going to do all the maintainance work and how do we handle the republishing of the crates with the correct new repository URL?

panic-semihosting/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+    "The Cortex-M Team <cortex-m@teams.rust-embedded.org>",
+    "Jorge Aparicio <jorge@japaric.io>",
+]
+categories = ["no-std"]
+description = "Report panic messages to the host stderr using semihosting"
+documentation = "https://docs.rs/panic-semihosting"
+keywords = ["panic-handler", "panic-impl", "panic", "semihosting"]
+license = "MIT OR Apache-2.0"
+name = "panic-semihosting"
+repository = "https://github.com/rust-embedded/panic-semihosting"
```
adamgreig, line 12:
```suggestion
repository = "https://github.com/rust-embedded/cortex-m"
```

cortex-m-semihosting/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+    "The Cortex-M Team <cortex-m@teams.rust-embedded.org>",
+    "Jorge Aparicio <japaricious@gmail.com>",
+]
+description = "Semihosting for ARM Cortex-M processors"
+documentation = "https://docs.rs/cortex-m-semihosting"
+keywords = ["semihosting", "arm", "cortex-m"]
+license = "MIT OR Apache-2.0"
+name = "cortex-m-semihosting"
+readme = "README.md"
+repository = "https://github.com/rust-embedded/cortex-m-semihosting"
```
adamgreig, line 12:
```suggestion
repository = "https://github.com/rust-embedded/cortex-m"
```

cortex-m-semihosting/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,37 @@
...
+
+And if someone takes issue with something you said or did, resist the urge to be defensive. Just stop doing what it was they complained about and apologize. Even if you feel you were misinterpreted or unfairly accused, chances are good there was something you could've communicated better — remember that it's your responsibility to make your fellow Rustaceans comfortable. Everyone wants to get along and we are all here first and foremost because we want to talk about cool technology. You will find that people will be eager to assume good intent and forgive as long as you earn their trust.
+
+The enforcement policies listed above apply to all official embedded WG venues; including official IRC channels (#rust-embedded); GitHub repositories under rust-embedded; and all forums under rust-embedded.org (forum.rust-embedded.org).
+
+*Adapted from the [Node.js Policy on Trolling](http://blog.izs.me/post/30036893703/policy-on-trolling) as well as the [Contributor Covenant v1.3.0](https://www.contributor-covenant.org/version/1/3/0/).*
+
+[team]: https://github.com/rust-embedded/wg#the-cortex-m-team
```
adamgreig, line 37: Can we just delete this file? It already exists in the repository root. And the LICENCE files too in fact. Same in the panic-semihosting directory.

cortex-m-semihosting/README.md
```diff
@@ -0,0 +1,40 @@
...
+This crate is guaranteed to compile on stable Rust 1.33.0 and up. It *might*
+compile with older versions but that may change in any new patch release.
+
+## License
+
+Licensed under either of
+
+- Apache License, Version 2.0 ([LICENSE-APACHE](LICENSE-APACHE) or
+  http://www.apache.org/licenses/LICENSE-2.0)
+- MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
```
adamgreig, line 23: If we do delete the redundant licence files, these links would need updating to point to the file in the directory above, as far as I can tell that should still work even on crates.io though.

cortex-m-semihosting/README.md
```diff
@@ -0,0 +1,40 @@
...
+for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
+dual licensed as above, without any additional terms or conditions.
+
+## Code of Conduct
+
+Contribution to this crate is organized under the terms of the [Rust Code of
+Conduct][CoC], the maintainer of this crate, the [Cortex-M team][team], promises
+to intervene to uphold that code of conduct.
+
+[CoC]: CODE_OF_CONDUCT.md
```
adamgreig, line 39: This link would also need updating (and in panic-semihosting's README).

adamgreig (changes requested):
Ok, here's a proposal based on the meeting yesterday:

* [x] Update links as suggested in this review
* [x] Fix merge errors by rebuilding precompiled libs
* [x] Merge this PR
* [x] Apply https://github.com/rust-embedded/cortex-m-semihosting/pull/41/files in a new PR
* [x] Check for breakage and update cortex-m dependency in cortex-m-semihosting and panic-semihosting to allow cortex-m <0.8 instead of <0.7
* [x] Release cortex-m-semihosting 0.4.0 with those changes, publishing the new links and ensuring compatibility with the new cortex-m
* [x] Release panic-semihosting 0.5.4, same deal
* [x] Mark the two old repositories as archived
* [x] Merge https://github.com/rust-embedded/panic-itm into this repository as well
* [x] Update links, bump cortex-m dependency to allow 0.7.0, release v0.4.2
* [x] Mark panic-itm repository as archived
* [x] Release cortex-m 0.7

We could also consider:
* [ ] Apply formatting from https://github.com/rust-embedded/cortex-m-semihosting/pull/54 but to all of cortex-m; this is a large but non-breaking change that we could probably land before 0.7 and might want to do before 1.0.

adamgreig:
@jonas-schievink any thoughts on my review comments? If we can get those ticked off and rebuild the binary files I think we're clear to merge this and start on the rest of the checklist.

adamgreig (approved):

adamgreig:
bors merge

PR #264, opened 2020-09-03
Tidy up some inline asm and add compiler fences where appropriate

adamgreig:
This PR updates the inline asm:

* Use compiler-assigned registers instead of specifying r0/r1/r2
* Write multi-line asm as multiple string literals, with normal Rust comments outside the strings
* Add ISB after writing to CONTROL as per ARM architectural requirements (see eg app note 321). As far as I can see no other requirements from AN321 apply here.
* Add compiler fences around enabling and disabling interrupts
    * No runtime barriers are required, but the compiler fences ensure the compiler won't reorder instructions around these operations, which would break critical section soundness.
* Add compiler fences around DMB, DSB, ISB to align compiler behaviour with the barrier runtime behaviour.
* Add compiler fences after the cache enable routines and writing to CONTROL since those routines include an ISB instruction.

Open to feedback on whether more or fewer fences are necessary; I've thought about these a bit but I think it's a tricky subject. I think in general the FFI-esque treatment of the new `asm!` block probably does most of what we need, but I'm told LLVM may still reorder instructions around FFI calls, which we really don't want to happen here.


therealprof:
Nice one. Looks good on first glance but I want to give people some time to look it over.

therealprof:
Hm, superweird that some archives are shrinking massively in size. These functions are always compiled with optimisation? Otherwise the `compiler_fence`s are going to be really bad...

adamgreig:
I don't understand the GitHub percentages (I think it's "number of lines" somehow?); the files are all around 14k-25kB; the changes are all less than 1kB. I don't understand why they're changing this much still though, the ones I inspected manually had some different debug strings in the ELF but otherwise no obvious clues.

adamgreig:
The xtask build generates the staticlib and LTO libs with optimisations enabled. If you're using the `inline-asm` feature you directly use this code, in which case optimisation depends on your own build.

asm/inline.rs
```diff
@@ -20,45 +22,58 @@ pub unsafe fn __control_r() -> u32 {
 
 #[inline(always)]
 pub unsafe fn __control_w(w: u32) {
-    asm!("msr CONTROL, {}", in(reg) w);
+    // ISB is required after writing to CONTROL,
+    // per ARM architectural requirements (see Application Note 321).
+    asm!("msr CONTROL, {}", "isb", in(reg) w);
+
+    // Ensure instructions are not reordered around the CONTROL update.
+    compiler_fence(Ordering::SeqCst);
```
jonas-schievink, line 30: This only affects *memory* instructions. Probably a good idea to clarify that in the comments.

asm/inline.rs
```diff
@@ -20,45 +22,58 @@ pub unsafe fn __control_r() -> u32 {
 
 #[inline(always)]
 pub unsafe fn __control_w(w: u32) {
-    asm!("msr CONTROL, {}", in(reg) w);
+    // ISB is required after writing to CONTROL,
+    // per ARM architectural requirements (see Application Note 321).
+    asm!("msr CONTROL, {}", "isb", in(reg) w);
```
jonas-schievink, line 27: Having each instruction in its own line is much more readable IMO

therealprof:
> The xtask build generates the staticlib and LTO libs with optimisations enabled. If you're using the `inline-asm` feature you directly use this code, in which case optimisation depends on your own build.

I wonder whether that's the right play here or whether we should place the fences ourselves, rather than using `compiler_fence`. It doesn't sound great to get additional panicking paths in dev mode by using those primitives.

adamgreig:
Without the fences the functions are potentially unsound but they are marked unsafe, so we _could_ just document the fence requirement in each relevant method's safety requirements. It's a bit of a chore though and it means users _must_ put those fences in themselves, where they're subject to the exact same issues. At least having them here means that for most users not using `inline-asm`, they get the optimised build of these methods even when they're in debug mode themselves.

Having said that, it's not clear to me how the fence here makes a difference to a caller of either the staticlib or LTO lib, which presumably only sees the FFI call, which also suggests to me that LLVM really shouldn't dare reorder instructions around an FFI call.

therealprof (changes requested):
Oh, and as usual, this is missing a CHANGELOG entry. ;)

asm/inline.rs
```diff
@@ -20,45 +22,58 @@ pub unsafe fn __control_r() -> u32 {
...
+        in(reg) cyc
+    );
 }
 
-// FIXME do we need compiler fences here or should we expect them in the caller?
 #[inline(always)]
 pub unsafe fn __dmb() {
-    asm!("dmb 0xF");
+    asm!("dmb");
+    compiler_fence(Ordering::AcqRel);
```
thalesfragoso, line 64: > It ensures that all explicit memory accesses that appear in program order before the DMB instruction are observed before any explicit memory accesses that appear in program order after the DMB instruction.

Shouldn't this be `SeqCst` ? Consider the following case:

```
read - 1
-> 3 and 4 can be moved to here
write - 2

dmb
AcqRel

read - 3
-> 1 and 2 can be moved to here
write - 4
```

Is my reasoning correct ?

asm/inline.rs
adamgreig: I've done so, but will the `asm!` block's default memory clobber ensure no memory accesses can be reordered around the block anyway? Will the compiler fence even survive being turned into an FFI call?

I've added these fences on the understanding that the compiler might otherwise reorder instructions around the `asm` call, but I don't fully see what the fence would prevent that's not prevented by the `asm!` block and/or FFI call (with or without LTO inlining).

adamgreig:
> Oh, and as usual, this is missing a CHANGELOG entry. ;)

There's no CHANGELOG entry for the new asm stuff at all, and it doesn't change the user-facing API (in other words it's an implementation detail). I've added a very brief description to the CHANGELOG about it.

asm/inline.rs
adamgreig: Fair, it stood out a bit as the only AcqRel anyway. I've updated it to SeqCst for now.

asm/inline.rs
adamgreig: Makes sense, I've reformatted as such.

therealprof:
> There's no CHANGELOG entry for the new asm stuff at all,

I agree, but that doesn't make it any better.

> and it doesn't change the user-facing API (in other words it's an implementation detail).

Not quite. It can now be linked with via the LTO linker plugin.

therealprof (approved):
LGTM

adamgreig:
> Not quite. It can now be linked with via the LTO linker plugin.

Good point, I've added a specific note about that to the CHANGELOG too.

therealprof:
Hm, weird. Why is it complaining about the blobs not being up-to-date?

bors try

jonas-schievink:
Because 5d5e15181f4d4eeacb9876a514f7359527f9b361 didn't update the blobs

therealprof (approved):
@jonas-schievink Ah, thanks!

I guess all comments have been sufficiently addressed so I'm inclined to get this on the road then?

Anyone want to give it another nod and pull the trigger?

adamgreig:
I'm away from my computer for a few days but if anyone else wants to rebuild the blobs and push to the branch please go ahead!

therealprof:
> I'm away from my computer for a few days but if anyone else wants to rebuild the blobs and push to the branch please go ahead!

Already done.

adamgreig:
bors r=therealprof

PR #268, opened 2020-10-02
Add some Armv8-M assembly routines

hug-dev:
Adds access to `MSP_NS` and the `BXNS` instruction.
Also adds `__dsb` which was missing.

Executed `cargo xtask assemble` and pushed everything, I hope that's enough.


src/register/msp.rs
```diff
@@ -11,3 +11,19 @@ pub fn read() -> u32 {
 pub unsafe fn write(bits: u32) {
     call_asm!(__msp_w(bits: u32));
 }
+
+/// Reads the Non-Secure CPU register from Secure state.
+/// Executing this function in Non-Secure state will return zeroes.
```
jonas-schievink, line 16:
```suggestion
///
/// Executing this function in Non-Secure state will return zeroes.
```

src/register/msp.rs
```diff
@@ -11,3 +11,19 @@ pub fn read() -> u32 {
...
+#[cfg(armv8m)]
+#[inline]
+pub fn read_ns() -> u32 {
+    call_asm!(__msp_ns_r() -> u32)
+}
+
+/// Writes `bits` to the Non-Secure CPU register from Secure state.
+/// Executing this function in Non-Secure state will be ignored.
```
jonas-schievink, line 24:
```suggestion
///
/// Executing this function in Non-Secure state will be ignored.
```

jonas-schievink (commented):
LGTM, added a small rustdoc improvement

hug-dev:
Done both :)

jonas-schievink (approved):
bors r+

PR #273, opened 2020-10-20
c-m-sh: fix missing prebuilt binaries. closes #271

### LINKED ISSUE #271: Latest release of cortex-m-semihosting seems broken?
After the 0.4.0 update of cortex-m-semihosting we are seeing breakage in RTIC CI pipeline as follows:

```console
error: failed to run custom build command for `cortex-m-semihosting v0.4.0`

Caused by:
  process didn't exit successfully: `/home/emifre/Git/rtic/cortex-m-rtic/target/debug/build/cortex-m-semihosting-6c5d0dd3173d1553/build-script-build` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/emifre/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-semihosting-0.4.0/build.rs:15:14
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
```

No idea what is actually wrong here, but it seems something broke in the latest release?

adamgreig:
Adding a symlink to `../bin` allows `cargo package` to include a copy of the binary files in the resulting tarball, which means the build script is able to link them as normal.

* We include the `-lto.a` files, but don't have an LTO_PLUGIN feature and so never use them
* While this should work for all platforms to use from crates.io, it's not clear what e.g. Windows will do with the symlink when trying to use this as a path/patch dependency directly from the git repository.

If it seems useful I could probably add a LTO feature the same as cortex-m, but I'd rather wait until we can have c-m-sh just directly use cortex-m's `syscall` method and then users can enable that feature in cortex-m themselves instead.

If anyone has this set up on Windows and is able to see if you can still use this repository as a path dependency, that would be great!


cortex-m-semihosting/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.4.1] - 2020-10-20
```
therealprof, line 8: What did we say we'd do about those links again? Maintain them with special tags? If not we can just give them up but then this shouldn't be a link.

therealprof (changes requested):

therealprof (commented):
Other than this one nit looks good to me. Did you want to wait for (Windows) people to chime in before or just do it and fix the fallout (if any) later?

cortex-m-semihosting/CHANGELOG.md
adamgreig: Thanks for catching that, I had intended for it to be a link and I'll make the tag upon release. I've pushed an update.

adamgreig:
I'd rather get this fix in now and released, if that sounds OK. I'm not sure if we can do _anything_ to properly fix this if it doesn't work with symlinks on Windows; we might have to suggest people copy the artifacts from `../bin/` into the `cortex-m-semihosting/bin` directory when using the repository as a local dependency.

Once we swap to using the `syscall` method from cortex-m directly this problem goes away.

jonas-schievink:
I mean, we could also just copy them there for 1 release and delete them again

therealprof (approved):
LGTM

adamgreig:
If we delete them again is that any better for Windows users? Perhaps the xtask build script should copy them into the c-m-sh subdirectory for us, and check them?

adamgreig:
I think let's merge this to get it fixed for now, and if it turns out to not work for Windows developers we can investigate having xtask copy the binaries into the semihosting project. Since the potential issue only affects people using the git repository, we can also fix this later without requiring another release.

I guess for cortex-m 1.0 we can release semihosting 0.5 (or just integrate it into cortex-m 1.0, who knows) which uses the cortex-m pre-built binaries.

bors merge

PR #274, opened 2020-10-20
Drop AT&T syntax from inline asm

### LINKED ISSUE #267: inline-asm feature causes compile errors on cortex-m0 (samd21)
Using version 0.6.3 with feature `inline-asm` enabled causes compile errors:

``` shell
error: invalid instruction, any one of the following would fix this:
  --> /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.6.3/src/asm.rs:39:13
   |
39 |             llvm_asm!("1:
   |             ^
   |
note: instantiated into assembly here
  --> <inline asm>:3:19
   |
3  |                   subs r1, $1
   |                   ^

note: instruction requires: thumb2
  --> /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.6.3/src/asm.rs:39:13
   |
39 |             llvm_asm!("1:
   |             ^
   |
note: instantiated into assembly here
  --> <inline asm>:3:19
   |
3  |                   subs r1, $1
   |                   ^

note: invalid operand for instruction
  --> /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.6.3/src/asm.rs:39:13
   |
39 |             llvm_asm!("1:
   |             ^
   |
note: instantiated into assembly here
  --> <inline asm>:3:28
   |
3  |                   subs r1, $1
   |                            ^

note: operand must be an immediate in the range [0,255]
  --> /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.6.3/src/asm.rs:39:13
   |
39 |             llvm_asm!("1:
   |             ^
   |
note: instantiated into assembly here
  --> <inline asm>:3:28
   |
3  |                   subs r1, $1
   |                            ^

note: operand must be a register in range [r0, r7]
  --> /home/johnny/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.6.3/src/asm.rs:39:13
   |
39 |             llvm_asm!("1:
   |             ^
   |
note: instantiated into assembly here
  --> <inline asm>:3:28
   |
3  |                   subs r1, $1
   |                            ^

error: aborting due to previous error

error: could not compile `atsamd-hal`
```

### LINKED ISSUE #272: asm::delay with inline-asm feature is broken on latest nightly
`asm::delay` hangs up on `rustc 1.49.0-nightly (8dae8cdcc 2020-10-12)`. Target is thumbv7em-none-eabihf.

The compiler produces following binary. It looks like the operand of `subs` is screwed up.

<details>
<summary>objdump</summary>

```
0800023c <_ZN8cortex_m3asm5delay17h6801139b6c4b4672E>:
 800023c: 80 b5         push    {r7, lr}
 800023e: 6f 46         mov     r7, sp
 8000240: 84 b0         sub     sp, #16
 8000242: 03 90         str     r0, [sp, #12]
 8000244: 01 21         movs    r1, #1
 8000246: 01 eb 90 00   add.w   r0, r1, r0, lsr #2
 800024a: 00 21         movs    r1, #0
 800024c: 00 29         cmp     r1, #0
 800024e: 01 90         str     r0, [sp, #4]
 8000250: 09 d1         bne     #18 <_ZN8cortex_m3asm5delay17h6801139b6c4b4672E+0x2e>
 8000252: ff e7         b       #-2 <_ZN8cortex_m3asm5delay17h6801139b6c4b4672E+0x1c>
 8000254: 01 98         ldr     r0, [sp, #4]
 8000256: 02 90         str     r0, [sp, #8]
 8000258: 02 99         ldr     r1, [sp, #8]

 800025a: 00 bf         nop
 800025c: b1 39         subs    r1, #177 <---- !?
 800025e: fc d1         bne     #-8 <_ZN8cortex_m3asm5delay17h6801139b6c4b4672E+0x22>

 8000260: 02 91         str     r1, [sp, #8]
 8000262: 04 b0         add     sp, #16
 8000264: 80 bd         pop     {r7, pc}
 8000266: 41 f2 b0 30   movw    r0, #5040
 800026a: c0 f6 00 00   movt    r0, #2048
 800026e: 41 f2 94 32   movw    r2, #5012
 8000272: c0 f6 00 02   movt    r2, #2048
 8000276: 1c 21         movs    r1, #28
 8000278: 00 f0 1f fd   bl      #2622
 800027c: fe de         trap
```
</details>

Please publish v0.6.4 (or 0.7?) with #262 included, if possible.

thalesfragoso:
This was breaking compilation on thumbv6 targets and miscompiling on thumbv7.

Fixes #272 
Fixes #267

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,22 +26,22 @@ matrix:
 
     # MSRV
     - env: TARGET=thumbv6m-none-eabi
-      rust: 1.31.0
+      rust: 1.36.0
       if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     # MSRV
     - env: TARGET=thumbv7m-none-eabi
-      rust: 1.31.0
+      rust: 1.36.0
       if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     # MSRV
     - env: TARGET=thumbv7em-none-eabi
-      rust: 1.31.0
+      rust: 1.36.0
       if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     # MSRV
     - env: TARGET=thumbv7em-none-eabihf
-      rust: 1.31.0
+      rust: 1.36.0
       if: (branch = staging OR branch = trying) OR (type = pull_request AND branch = master)
 
     - env: TARGET=thumbv6m-none-eabi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- MSRV bumped to 1.36.0 due to `aligned` dependency.
+
+### Fixed
+
+- Drop AT&T syntax from inline asm, which was causing miscompilations with newer versions of the compiler.
+
 ## [v0.6.3] - 2020-07-20
 
 ### Added
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -38,7 +38,7 @@ pub fn delay(_n: u32) {
         () => unsafe {
             llvm_asm!("1:
                   nop
-                  subs $0, $$1
+                  subs $0, #1
                   bne.n 1b"
                  : "+r"(_n / 4 + 1)
                  :
@@ -80,7 +80,6 @@ pub fn nop() {
         () => unimplemented!(),
     }
 }
-
 
 /// Generate an Undefined Instruction exception.
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,7 @@
 //!
 //! # Minimum Supported Rust Version (MSRV)
 //!
-//! This crate is guaranteed to compile on stable Rust 1.31 and up. It *might*
+//! This crate is guaranteed to compile on stable Rust 1.36 and up. It *might*
 //! compile with older versions but that may change in any new patch release.
 
 #![cfg_attr(feature = "inline-asm", feature(llvm_asm))]
@@ -34,7 +34,6 @@
 #![no_std]
 #![allow(clippy::identity_op)]
 #![allow(clippy::missing_safety_doc)]
-
 // This makes clippy warn about public functions which are not #[inline].
 //
 // Almost all functions in this crate result in trivial or even no assembly.


jonas-schievink (approved):
bors r+

thalesfragoso:
Hmm, `aligned` depends on `as-slice` that depends on `generic-array` that uses `MaybeUninit`, which breaks our MSRV.

jonas-schievink:
Just bump it, master doesn't use those anymore

thalesfragoso:
> Just bump it, master doesn't use those anymore

We can't release 0.6.4 if we do that. I will try to get rid of aligned, should be easy.

jonas-schievink:
That's a breaking change, since we use it in public APIs. Why can't we bump the MSRV?

thalesfragoso:
> //! # Minimum Supported Rust Version (MSRV)
//!
//! This crate is guaranteed to compile on stable Rust 1.31 and up. It *might*
//! compile with older versions but that may change in any new patch release.

But you are right about removing `aligned`, hmmm

jonas-schievink:
I don't think the intention was to make that part of our semver guarantees. We cannot give more guarantees than our dependencies there.

adamgreig:
At least in principle we [do require](https://github.com/rust-embedded/wg/blob/master/ops/msrv.md) a semver bump for MSRV changes, which would prevent releasing 0.6.4 with an MSRV above 1.31.

Can we require an older version of `aligned`?

thalesfragoso:
> Can we require an older version of aligned?

Wouldn't that also be breaking ?

adamgreig:
Yes, I suppose it would. Shame for us that nothing in the aligned -> generic_array -> as_slice link made a semver bump when increasing their MSRV, but probably points to our current policy being untenable in general. Sounds like bumping the MSRV for 0.6.4 is probably the least bad option?

thalesfragoso:
> Sounds like bumping the MSRV for 0.6.4 is probably the least bad option?

I think so, should I change the version in the docs or just remove it entirely ? I think the later is option is probably better, since we don't control all of our dependencies.

adamgreig:
Throwing away the MSRV entirely seems a bit more drastic for a bug-fix patch release, even if it might become wrong if our dependencies increase it again. What MSRV will we need for this release?

thalesfragoso:
> Throwing away the MSRV entirely seems a bit more drastic for a bug-fix patch release, even if it might become wrong if our dependencies increase it again. What MSRV will we need for this release?

1.36. I think it doesn't make much sense to say that this crate is guaranteed to compile with some specific version if we can't guarantee that...

therealprof:
If our dependencies bumped the MSRV without following semver best practises, doesn't that automatically mean they already made the decision for us (unless we want to pin a specific version of that dependency)?

thalesfragoso:
Well, the `aligned` crate doesn't have a MSRV to start with, we can't blame them. The problem was that we added a crate with an incompatible MSRV. It seems to be a bit hard to include MSRV in semver, unless you control all your dependencies.

adamgreig:
> If our dependencies bumped the MSRV without following semver best practises, doesn't that automatically mean they already made the decision for us (unless we want to pin a specific version of that dependency)?

I don't think it's well established what the semver "best practices" even are around MSRV, or at least I couldn't find anything when we last worried about this. It's possible we both have sensible but incompatible policies.

therealprof:
Well, I was thinking about the general semver best practises (unrelated to Rust) but actually they really only seem to bother about API surface and not other factors like MSRV so that point is moot.

adamgreig:
To get this fix in, shall we bump our reported MSRV for now (and for this 0.6.4 release) to 1.36, leave it in the documentation, and we can spend some more time considering what to do about MSRV generally afterwards? If a user still needs a lower MSRV, at least nothing in _our_ crate stops it; they could always specify an earlier version of `aligned` in their own `Cargo.toml` and which is still compatible with `cortex-m` and should allow the build to continue on earlier rustc.

thalesfragoso:
> To get this fix in, shall we bump our reported MSRV for now (and for this 0.6.4 release) to 1.36, leave it in the documentation, and we can spend some more time considering what to do about MSRV generally afterwards? If a user still needs a lower MSRV, at least nothing in _our_ crate stops it; they could always specify an earlier version of `aligned` in their own `Cargo.toml` and which is still compatible with `cortex-m` and should allow the build to continue on earlier rustc.

I don't feel very comfortable leaving the MSRV notice as it's now for 0.6.4, maybe change the wording ? We can bring it back on 0.7, given that we can control our dependencies (we already dropped `aligned`) or at least ask them for setting a MSRV, if we decide on going forward with having a MSRV aware semver policy.

thalesfragoso:
Ok, I don't think this fix should be blocked by this, so I updated the MSRV and left things as it was. I can change more things if it's required, leaving this decision to you folks.

thalesfragoso:
bors try

therealprof (approved):
LGTM, too.

bors r+

adamgreig:
> I don't feel very comfortable leaving the MSRV notice as it's now for 0.6.4, maybe change the wording ? We can bring it back on 0.7, given that we can control our dependencies (we already dropped `aligned`) or at least ask them for setting a MSRV. That is, if we decide on going forward with having a MSRV aware semver policy.

Sorry, I didn't mean to leave this sitting. I still think it makes sense to get 0.6.4 out now, but I think overall you're right about the MSRV, as this example clearly illustrates. I still think there's value in stating an MSRV at least so that we're aware of when we change it and can consider if a PR merits a large bump, especially for a crate like cortex-m which is going to be quite widely used in embedded. Maybe it doesn't make sense for bumping the MSRV to be a semver breaking change, though, and maybe we will need to be more careful of dependencies too.

PR #275, opened 2020-10-26
Prepare 0.6.4

adamgreig:
Includes #274 to fix recent issues with inline-asm on nightly.

PR #276, opened 2020-10-26
Set up GitHub Actions

jonas-schievink:


jonas-schievink:
bors try

jonas-schievink:
bors try

jonas-schievink:
bors try-

jonas-schievink:
bors try

jonas-schievink:
@adamgreig ready for review :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+        # All generated code should be running on stable now
+        rust: [stable]
+
+        include:
+          # Test MSRV
+          - rust: 1.38.0
+
+          # Test nightly but don't fail
+          - rust: nightly
+            experimental: true
```
adamgreig, line 22: By itself this won't do anything, `experimental` isn't magic (I think `wg/ops/post-transfer.md` is wrong here). You need to set `continue-on-error` to [match against your flag](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), and (probably?) set `experimental` to `false` for the other jobs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - name: Run tests
+        run: cargo test --all
+
+# FIXME: test on macOS and Windows
```
adamgreig, line 33: Unlike the Travis config this doesn't have a cache enabled, but since cortex-m has few and simple dependencies maybe there's not much advantage.

adamgreig (changes requested):
Thank you so much!! Looks good, couple of nits. Only shame is that merging this to master won't help #275 finally merge, sigh.

.github/workflows/ci.yml
jonas-schievink: Yeah, I want to avoid dealing with caching since it can be quite unreliable

adamgreig: Fine with me.

jonas-schievink: CI time went down to ~2 minutes with this. Looks like we don't need caching for now :smile: 

(maybe this will change once we import cortex-m-rt, but I doubt it)

.github/workflows/ci.yml
jonas-schievink: Yeah, that's odd, it was missing from the template. I opened https://github.com/rust-embedded/wg/pull/518 to fix that.

adamgreig:
Ok, looks good to me. My only remaining concern is a policy one around whether we want to enforce rustfmt by failing PRs that don't have it run; we don't currently do this on any (I think?) wg repos. I know it's a tired fight but I at least don't want it to happen _de facto_. Sadly there's presumably not many good options...

* allowing CI to pass despite rustfmt failure makes it basically meaningless
* no current way to have bors or github apply rustfmt for us
* not doing it presumably means users with auto-rustfmt will keep adding accidental formatting changes to their PRs until it's all rustfmt-clean anyway

I don't auto-rustfmt so I'm sort of meh about the whole situation but I don't think I can muster any compelling arguments against.

On another note, I see the Clippy run fails because the PR is from a fork; do we expect that will work after this PR is merged, or should we change the clippy template too? I know GHA supports running the base branch's CI workflow on the PR's git commit, which might allow clippy annotations for PRs-from-forks, but maybe this will already work since it's what we're using elsewhere (or maybe it doesn't work elsewhere either, and we don't notice when making non-fork PRs?).

adamgreig (approved):

jonas-schievink:
> I see the Clippy run fails because the PR is from a fork; do we expect that will work after this PR is merged, or should we change the clippy template too?

This unfortunately fundamentally doesn't work on GitHub Actions. I think the Clippy action will write the clippy output to the console instead though.

adamgreig:
> This unfortunately fundamentally doesn't work on GitHub Actions. I think the Clippy action will write the clippy output to the console instead though.

Ah, I see for example over on probe-rs that's what we get ([e.g.](https://github.com/probe-rs/probe-rs/pull/401/checks?check_run_id=1306010537#step:6:603)). I wonder if it's possible if we use [`pull_request_target`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) instead of `pull_request`, which gives access to secrets to the action and might allow clippy to post annotations. I've used it in stm32-rs to be able to push generated files to another repository triggered by PRs from forks. Anyway we can experiment with it later, there's no need for it to block this.

adamgreig:
Oh, we already do rustfmt on cortex-m-rt, so that fight has been and gone. Let's do this!

bors merge

jonas-schievink:
> I wonder if it's possible if we use [`pull_request_target`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) instead of `pull_request`, which gives access to secrets to the action and might allow clippy to post annotations.

Oh, sounds like they added this specifically for clippy-style actions, neat!

PR #277, opened 2020-10-26
ci: run Clippy on pull_request_target

jonas-schievink:


adamgreig:
I can't comment on the relevant line because it's not close enough to the parts that changed (thanks, github???) but I don't think the checkout action in clippy.yml will work; it would check out the base/target branch instead of the PR branch.

I used something like this in stm32-rs:

```yaml
- name: Checkout PR
  uses: actions/checkout@v2
  with:
    ref: refs/pull/${{ github.event.number }}/head
```

adamgreig (approved):
Can't test until it's merged so let's give it a go.

adamgreig:
bors merge

PR #285, opened 2020-11-04
`xtask assemble` fixes

### LINKED ISSUE #284: Panic handler removal does not work
`trim_panic_handler` checks for a section containing the string `cortex_m_asm_panic`, but we never set that string anywhere.

I guess I meant to put `#[link_section]` on this function, but never did?

https://github.com/rust-embedded/cortex-m/blob/fd6a9dbdab0a9c1b377c04a2820324cd8337706f/asm/lib.rs#L123-L139

jonas-schievink:
Closes #284


thalesfragoso (approved):
Great, thanks!

bors r+

PR #292, opened 2020-11-14
Prepare for panic-itm v0.4.2 supporting cortex-m v0.7

adamgreig:


panic-itm/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.4.2] - 2020-11-14
+
+- Support cortex-m v0.7.0
```
jonas-schievink, line 12: This is a breaking change, because cortex-m 0.7 is incompatible with 0.6, right? So silently updating would break your build.

adamgreig: But panic-itm would still support old cortex-m, so a build that's currently using cortex-m v0.6 should continue doing so?

jonas-schievink: Ah, right

jonas-schievink (approved):
bors r+

PR #297, opened 2020-11-25
Deprecate msp::write

jonas-schievink:
The function can not be used from Rust code, because it destroys the stack frame.

Users must use an assembly block (inline or out-of-line) that both sets MSP and performs a method call.

### DIFF
--- a/src/register/msp.rs
+++ b/src/register/msp.rs
@@ -8,6 +8,7 @@ pub fn read() -> u32 {
 
 /// Writes `bits` to the CPU register
 #[inline]
+#[deprecated = "calling this function invokes Undefined Behavior"]
 pub unsafe fn write(bits: u32) {
     call_asm!(__msp_w(bits: u32));
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - New assembly methods `asm::semihosting_syscall`, `asm::bootstrap`, and
   `asm::bootload`.
 
+### Changed
+
+- `msp::write` has been deprecated in favor of `asm::bootstrap`. It was not
+  possible to use `msp::write` without causing Undefined Behavior, so all
+  existing users are encouraged to migrate.
+
 ## [v0.7.0] - 2020-11-09
 
 ### Added


adamgreig:
I think we need to provide a usable alternative that works on stable before we deprecate this; I know I use it a bunch for either jumping to bootloaders or jumping to application code, and (as we keep finding out with cortex-m/-rt) it's a fair pain to include an assembly blob without using unstable inline-asm.

Perhaps that just needs to be the function we talked about that reads an SP and RV from a vector table and does the MSP write and jump, which is of course unsafe but might at least be sound so long as you give it the address of a valid vector table. That said, couldn't the same be true here so long as the user ensures immediately after updating MSP they jump into their new (divergent) function?

jonas-schievink:
I think having a function like `fn bootstrap(stack_pointer: *mut u32, entry: fn() -> !)` would be enough for that, right?

One problem is that that doesn't allow passing parameters, so maybe `entry` should take 4 `u32` arguments?

Do we want to provide that in cortex-m or should all the assembly stubs be put into a cortex-m-asm crate, to be useful outside of bare-metal targets? (now that I think about it, such a crate would be a lot easier to push to 1.0 than the entirety of cortex-m)

adamgreig:
I think just `fn bootload(vector_table: *const u32)` would suffice for bootloading; the first entry in the table is always the new stack pointer and the second is always the new reset vector, in cortex-m. Saves the user having to transmute a raw pointer to a `fn() -> !` themselves too. I guess the reason to pass arguments (or jump to something that's not a vector table) is OS implementations, perhaps they can have a separate method? I'm not as clear what they'd need, but it wouldn't hurt to support them too.

Putting these and the other asm routines into their own cortex-m-asm crate sounds great.

jonas-schievink:
RTOSes probably want to use PSP instead of MSP anyways, so +1 to just providing a bootloader-only method for now.

adamgreig:
The great thing about cortex-m-asm is even after 1.0 almost anything we'd want to do is add new methods, which aren't breaking changes anyway.

therealprof:
Sounds like a good plan. I'd prefer the name `bootstrap` over `bootload` in case we're bikeshedding.

src/register/msp.rs
```diff
@@ -8,6 +8,7 @@ pub fn read() -> u32 {
 
 /// Writes `bits` to the CPU register
 #[inline]
+#[deprecated = "calling this function invokes Undefined Behavior"]
```
adamgreig, line 11:
```suggestion
#[deprecated = "calling this function invokes Undefined Behavior, consider asm::bootstrap as an alternative"]
```

adamgreig:
I think we're good to merge this now. Could you add a changelog entry please?

adamgreig (approved):
bors r+

adamgreig:
(this looks fine to me and I think having the note about bootstrap in the changelog is OK, but just checking you deliberately reverted the deprecation message rather than a force-push accident?)

jonas-schievink:
> (this looks fine to me and I think having the note about bootstrap in the changelog is OK, but just checking you deliberately reverted the deprecation message rather than a force-push accident?)

Oh, no, that was an accident. I didn't pull the commit fixing it.

PR #298, opened 2020-12-02
Try scheduling CI runs

adamgreig:
This is a first go at running CI jobs on a scheduled basis. This job will run at 7pm UTC today (Wednesday), run our tests, and, in this case, fail. On failure it should then open a new issue on the repository with a link to the failed CI job.

We can bikeshed the schedule (this is just to get it to run soonish), and obviously I'll do another PR to remove the failing step once we've checked it works.

Please don't merge if we miss 7pm weds; I'll update the schedule first.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,0 +1,31 @@
+on:
+  schedule:
+    # Run every week at 7pm UTC Wednesday.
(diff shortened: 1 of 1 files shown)


thalesfragoso (approved):
Thanks, let's test this.

bors r+

PR #300, opened 2020-12-02
Update c-m-sh changelog for 0.3.6

adamgreig:
Copied from https://github.com/rust-embedded/cortex-m-semihosting/pull/61


jonas-schievink (approved):
bors r+

seems like we might need to yank this and make it 0.4 though: https://github.com/rust-embedded/cortex-m/issues/301

PR #305, opened 2020-12-08
Add required 'labels' field to scheduled CI, update next run date

adamgreig:
See also #298. This should fail at 10pm today _and open an issue this time_, then we'll update it to remove the intentional failure and perhaps set the scheduled date to something more useful.


therealprof (approved):
LGTM

bors r+

PR #307, opened 2020-12-09
Finish setting up scheduled CI job

adamgreig:
The scheduled CI seems to have worked (#306), so this PR removes the forced failure and sets the time to a hopefully sensible saturday morning, though I suppose it doesn't really matter that much what time of week it is.


jonas-schievink (approved):
bors r+

PR #312, opened 2021-01-07
Fix timing of asm-based delay implementation

jonas-schievink:
Should fix https://github.com/rust-embedded/cortex-m/issues/236

I've also moved the cycle adjustment into the assembly implementation, since it seems more appropriate there (and this will be useful if we make `cortex-m-asm` its own crate).


adamgreig (approved):
LGTM, thanks!

bors merge

PR #316, opened 2021-01-24
Prepare for v0.6.5 release

adamgreig:
From testing it seems like 0.6.5-alpha is OK, but it's really hard to use because you have to patch every dependency (HAL, PAC, etc) that's trying to use 0.6. I think the sooner we get the actual release out the better, now.


jonas-schievink (approved):
bors r+

adamgreig:
Released :tada:

PR #318, opened 2021-01-25
Use LTO instead of removing the panic handler manually. Closes #309

### LINKED ISSUE #309: Unable to debug application when using library version 0.7.0
First, I'am able to successfully flash the `STM32F303x` and run the application.
The issue I'm facing is attempting to debug the application with `GDB`/`OpenOCD`.

When issuing a `cargo run` I see the following messages (emphasis mine):

>Reading symbols from target/thumbv7em-none-eabihf/debug/app...
**D*W_FORM_strp pointing outside of .debug_str section*** [in module /Users/Me/Projects/embedded_template/target/thumbv7em-none-eabihf/debug/app]

>(No debugging symbols found in target/thumbv7em-none-eabihf/debug/app)

As a result I'am unable to set break points or use `layout src` in `GDB`.
This only occurs when using `cortex-m` version `0.7.0`.

If I use version `0.6.4` I'am able to successfully debug the application.

Aside from the `memory.x` file is there additional configuration needed for the newest version of the library?
Are specific versions of the library only compatible with specific `MCU`'s? 
Does the version of the `cortex-m` library require a specific version of `GDB`?

Below is my `memory.x` file if it helps:

    MEMORY
    {
        CCRAM : ORIGIN = 0x10000000, LENGTH = 8K
        FLASH : ORIGIN = 0x08000000, LENGTH = 256K
        RAM : ORIGIN = 0x20000000, LENGTH = 40K
    }
    _stack_start = ORIGIN(CCRAM) + LENGTH(CCRAM);

adamgreig:


jonas-schievink (approved):
bors r+

PR #317, opened 2021-01-25
Fix __delay clobber

jonas-schievink:

PR #311, opened 2021-01-07
Duplicate compiler fences around barrier instrs

jonas-schievink:
This should fix https://github.com/rust-embedded/cortex-m/issues/308

Issue #304, opened 2020-12-07
Cache maintenance operations require reference to self


ryan-summers:
Various cache maintenance-related operations require a `self` reference, which means that it is not possible for a HAL or other peripherals to clean/invalidate and/or perform other cache related operations without owning the `SCB`.

In the context of DMA, this can be a stumbling block, as DCACHE needs to be cleaned and invalidated before and after transfers (if it is enabled). 

Is it possible to refactor the cache maintenance functions to not require a `self` reference so that they can be called statically as `cortex_m::peripherals::scb::invalidate_dcache_by_address()`?

ryan-summers:
Related to #238

PR #320, opened 2021-01-25
Prepare for v0.7.1

adamgreig:
Includes:

* Deprecate msp::write #297
* New syscall and bootstrap ASM #299 
* More compiler fences #311 
* asm::delay timing fix #312 
* asm::delay clobber fix #317 
* LTO for ASM fix #318 

Doesn't include anything to help with #304 which might be nice to fix but can come in 0.7.2 and might need some time to think about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.7.1] - 2021-01-25
+
 ### Added
 
 - New assembly methods `asm::semihosting_syscall`, `asm::bootstrap`, and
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.7.1] - 2021-01-25
```
therealprof, line 10: I don't think we should release a 0.7.1 which doesn't list any signs of motivitation or changes. Are we going to yank 0.7.0? If so that is even more important...

therealprof (changes requested):

CHANGELOG.md
adamgreig: I don't understand what you mean exactly, the changelog contains everything that was previously "unreleased", i.e. has a bunch of added, deprecated, and fixed entries. What else should we include?

therealprof: Hm, weird; seems I was holding the PR wrong. Now I see it. 😅

therealprof (approved):
LGTM.

bors r+

CHANGELOG.md
therealprof: Okay, I was totally missing that #319 exists and was merged and the way the collapsed diff rendered led me to the conclusion that the changes were not there. Sorry about that.

PR #321, opened 2021-01-25
Fix `Stim` reexport on thumbv8m.base

jonas-schievink:


adamgreig:
We'll want a changelog update when we bump the version number. Is it worth re-enabling the test suite one way or another before releasing it, too?

jonas-schievink:
Yeah, updated the changelog. I don't want to touch 0.6.x again if possible, so backporting CI would just cost more time.

adamgreig (approved):
I've run `cargo check` across all the targets which is all the old CI script did anyway and it passes now, so hopefully this is the only issue. Do you want to push the release?

bors merge

jonas-schievink:
Published!

PR #322, opened 2021-01-26
Fix missing peripheral::itm export, prepare v0.6.7

adamgreig:
This fixes the issue where cortex-m 0.5.10 re-exports 0.6's peripheral::itm module; in principle it's possible other crates could have been using this module directly too (it is public) but in practice it doesn't contain anything useful (all the useful things are in the top-level itm module).

Should fix https://github.com/rust-embedded/discovery/issues/287.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


jonas-schievink (approved):
bors r+

therealprof (commented):
LGTM.

bors r+

PR #327, opened 2021-01-31
Prepare 0.5.11, removing linking to pre-built binaries.

adamgreig:
_I don't really like this. These semver hacks are getting out of hand and are mega cursed._

In https://github.com/rust-embedded/discovery/issues/288 it's reported that crates which use cortex-m 0.5 (such as our discovery book) have started to fail to build in some cases because of duplicate symbols such as `__cpsid` exported from the pre-built binaries.

It turns out in 0.5.10 we continue to link against a pre-built binary, but don't use it as the `asm` module is instead exported from `cortex_m_0_6` (i.e. 0.6.7, which itself exports it from 0.7.1).

What I don't understand is why this issue didn't appear before now, since it seems like 0.6.4 and below should have also exported these symbols and so should have also collided. I'd love some light on what's going on there.

This potential PR would release a 0.5.11 which just stops linking the pre-built binaries (and adds 0.5.9/10 to the changelog). I've tested that this change fixes the Discovery build, so it's one way we could fix this. Open to other, less cursed, suggestions too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.5.11] - 2021-01-31
+
+* This release stops linking the pre-built assembly files, as those methods are
+  instead provided by depending on cortex-m 0.6.x.
+
+## [v0.5.10] - 2019-04-29
+
+* This release ensures features set on this cortex-m crate are forwarded to
+  the included cortex-m 0.6.x crate.
+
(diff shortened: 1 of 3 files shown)


therealprof (approved):
I totally agree with those semver hacks being mega cursed. I think when 0.5 support becomes unattainable we should simply stop supporting them altogether instead of giving false hope that everything is under control.

bors r+

adamgreig:
This passed on the stable builds for all platforms, merging manually.

adamgreig:
Published 0.5.11.

PR #330, opened 2021-02-09
Fix rust-lld: error: undefined symbol: __syscall

ovidiusabou:
Encountered while building an example from stm32-rs/stm32f3xx-hal


jonas-schievink (approved):
bors r+

PR #334, opened 2021-02-19
Remove an extra function call in `primask::read`

yvt:
There doesn't seem to be any reason why the call to `__primask_r` should be wrapped by another function call. `read_raw` (the function being removed by this PR) is not `#[inline]`, so it did hurt performance when compiled without LTO.

Compare with `faultmask::read`:
https://github.com/rust-embedded/cortex-m/blob/7481f09d67aee7081d71cf6e042f2984c12a830f/src/register/faultmask.rs#L26-L35

### DIFF
--- a/src/register/primask.rs
+++ b/src/register/primask.rs
@@ -26,11 +26,7 @@ impl Primask {
 /// Reads the CPU register
 #[inline]
 pub fn read() -> Primask {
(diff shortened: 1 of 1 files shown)


jonas-schievink (approved):
bors r+

PR #328, opened 2021-02-08
Prevent underflow when calling delay(n) with n<2

ovidiusabou:
Calling delay(1) causes a very long wait (freeze) otherwise.
86cd463788b97ca3894936101c6cf4f3512f0fbc introduced this behaviour by
changing the cycle count from n / 4 + 1 to n / 2 which forces an
underflow when n<2.

PR #335, opened 2021-03-07
Prepare for v0.7.2 release of cortex-m

adamgreig:
We've had #328 merged for a while and it fixes a pretty annoying bug which is still in the wild. Can anything think of anything else to get in before a release? If there's nothing coming up (I think #282 needs some more discussion and #331 won't affect the published crate itself) we could get this released.


thalesfragoso (approved):
I agree, thanks!

bors r+

adamgreig:
[Published](https://crates.io/crates/cortex-m) :tada:

PR #341, opened 2021-04-25
Update warning about linker-plugin-lto not working with -Os/-Oz

Rahix:
Just noticed this while reading through the docs... The bug was fixed by @jonas-schievink in Rust 1.49 already, fortunately.  But as the MSRV is still 1.38 we cannot drop the warning entirely yet.


adamgreig (approved):
Thanks, good catch.

bors merge

PR #344, opened 2021-06-12
Remove dependence on embedded-hal for Delay

### LINKED ISSUE #343: Delay: Don't rely on EH traits
File in question: https://github.com/rust-embedded/cortex-m/blob/1a4e09646aa37c3408d69fe4452265f405400593/src/delay.rs#L32

Typically, HALs implement their own Delay. Given it only uses cortex-m peripheral, there's no reason HALs need to keep this; this `cortex-m` library includes a delay.

However, it's dependent on `embedded-hal::Delay`. Can we make the implementation native, and add EH as an optional add-on? This way, you don't need to import the trait, or include it in your project. I'll do a PR if approved.

David-OConnor:
Fixes #343

No functional changes. Allows you to use Delay as previously, or without including or importing `embedded-hal`.

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -29,7 +29,8 @@ impl Delay {
         self.syst
     }
 
-    fn _delay_us(&mut self, us: u32) {
+    /// Delay using the Cortex-M systick for a certain duration, µs.
+    pub fn delay_us(&mut self, us: u32) {
         let ticks = (us as u64) * (self.ahb_frequency as u64) / 1_000_000;
 
         let full_cycles = ticks >> 24;
@@ -54,6 +55,11 @@ impl Delay {
 
         self.syst.disable_counter();
     }
+
+    /// Delay using the Cortex-M systick for a certain duration, ms.
+    pub fn delay_ms(&mut self, ms: u32) {
+        self.delay_us(ms * 1_000);
+    }
 }
 
 impl DelayMs<u32> for Delay {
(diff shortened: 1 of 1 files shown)


David-OConnor:
I don't understand the inline bit, so LMK what you're looking for there.

src/delay.rs
```diff
@@ -54,17 +55,22 @@ impl Delay {
 
         self.syst.disable_counter();
     }
+
+    /// Delay using the Cortex-M systick for a certain duration, ms.
+    pub fn delay_ms(&mut self, ms: u32) {
+        self.delay_us(ms * 1_000);
```
Disasm, line 61: This implementation is correct not for all argument values. Please see the `delay_ms` implementation below.

David-OConnor:
Fixed, and took a stab at `#inline`.

Disasm:
I don't think it's a good idea to inline the main `delay_us` function, but maybe someone knows rustc better than me :) I made it a separate private function on purpose, so that all the other functions can be marked `#[inline]`.

David-OConnor:
Changed, and made the u32-conversion syntax consistent (went with `as` instead of `from`; can make all `from` if you want)

Disasm:
@David-OConnor I think `as` produced a warning somewhere that actually doesn't make sense in this specific case. `from` is better in general, because you can't unintentionally change types in a way that the program still compiles: if there is no `From` implementation it breaks, while `as` "works" even for `0x12345678u32 as u8`.

David-OConnor:
Done

src/delay.rs
```diff
@@ -29,8 +29,9 @@ impl Delay {
         self.syst
     }
 
-    fn _delay_us(&mut self, us: u32) {
-        let ticks = (us as u64) * (self.ahb_frequency as u64) / 1_000_000;
+    /// Delay using the Cortex-M systick for a certain duration, in µs.
+    pub fn delay_us(&mut self, us: u32) {
```
Disasm, line 33:
```suggestion
    #[allow(clippy::missing_inline_in_public_items)]
    pub fn delay_us(&mut self, us: u32) {
```

adamgreig (approved):
Thanks, LGTM!

bors merge

PR #337, opened 2021-03-19
Compile the `mpu::RegisterBlock` struct on all targets

dimpolo:
closes https://github.com/rust-embedded/cortex-m/issues/336

PR #345, opened 2021-06-25
Prepare for v0.7.3

adamgreig:
We've had #337 merged for a while and it seems like it would be helpful for users to get it released, and now #344 to go with it. No other changes since then (except a small doc change) so it should be an easy release.


therealprof (approved):
LGTM, thanks.

bors r+

PR #348, opened 2021-08-06
Fix slightly misleading examples.

newAM:
This is a bit of a nitpick; the examples provided for the peripherals use the cycle counter as an example, but it is not properly enabled, which is slightly misleading.  The examples only do `peripherals.DWT.enable_cycle_counter`, but this write may be ignored because tracing needs to be enabled first.


adamgreig (approved):
Thanks!

bors merge

adamgreig:
Hm, looks like bors forgot about this.

bors merge

PR #352, opened 2021-09-05
Add the Cortex-M7 TCM and cache access control registers.

rcls:
Add the Cortex-M7 TCM and cache access control registers.

These are documented in the Cortex-M7 generic user guide (ARM DUI 0646C). 

I'm not sure what feature gate these should be on - should I add a new one for Cortex-M7? Currently I have them on `not(armv6m)` - they do not appear to be in the ARMv7M architecture documentation, so I presume they are M7 specific.

### DIFF
--- a/src/peripheral/mod.rs
+++ b/src/peripheral/mod.rs
@@ -60,6 +60,8 @@ use core::ops;
 
 use crate::interrupt;
 
+#[cfg(cm7)]
+pub mod ac;
 #[cfg(not(armv6m))]
 pub mod cbp;
 pub mod cpuid;
@@ -91,6 +93,10 @@ mod test;
 #[allow(non_snake_case)]
 #[allow(clippy::manual_non_exhaustive)]
 pub struct Peripherals {
+    /// Cortex-M7 TCM and cache access control.
+    #[cfg(cm7)]
+    pub AC: AC,
+
     /// Cache and branch predictor maintenance operations.
     /// Not available on Armv6-M.
     pub CBP: CBP,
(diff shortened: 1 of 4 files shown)


thejpster:
This sounds like an example of where a feature in the processor core isn't tied to an architecture (e.g. `armv7e-m`) but instead to a specific processor design. As far as the Arm Armv7-M ARM is concerned, TCM is either "implementation defined" or "not available".

We already have a feature flag for `cm7-r0p1` to work around a specific silicon bug. Adding one for `cm7-tcm` doesn't seen unreasonable. Or just a generic `cm7` flag if the TCM function applies to all Cortex-M7 cores (where `cm7-r0p1` would depend on `cm7`).

rcls:
> Or just a generic `cm7` flag if the TCM function applies to all Cortex-M7 cores

I have added a generic `cm7` flag. (Looking closely, there are a few flags, e.g., the AXIM async bus fault, that appear to apply to all CM7s, not just those with TCM & caches.)

src/peripheral/mod.rs
```diff
@@ -91,6 +93,9 @@ mod test;
 #[allow(non_snake_case)]
 #[allow(clippy::manual_non_exhaustive)]
 pub struct Peripherals {
+    /// Cortex-M7 TCM and cache access control.
+    pub AC: AC,
```
adamgreig, line 97: Perhaps the presence of this field in the struct should be cfg-gated too? And then also the declaration of the struct itself.

rcls: Done

rcls:
Friendly "ping" on this, if anyone has a moment spare.

adamgreig (approved):
Thanks for the ping. All the values check out, looks good.

bors merge

PR #353, opened 2021-09-21
Bump dependencies.

hydra:
This PR has two commits.

1. one to bump the versions to the latest patch-level.
2. to allow latest patch release of a dependency - this reduces maintenance of this repo when the dependencies release a new version.

See also: 
* https://doc.rust-lang.org/cargo/reference/semver.html
* https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,10 +16,10 @@ edition = "2018"
 links = "cortex-m"  # prevent multiple versions of this crate to be linked together
 
 [dependencies]
-bare-metal = { version = "0.2.0", features = ["const-fn"] }
-volatile-register = "0.2.0"
+bare-metal = { version = "0.2.5", features = ["const-fn"] }
+volatile-register = "0.2.1"
 bitfield = "0.13.2"
-embedded-hal = "0.2.4"
+embedded-hal = "0.2.6"
 
 [features]
 cm7-r0p1 = []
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for this PR!

In Cargo, the caret is the [default operator if none are specified](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), so there's no need to add an explicit caret. Consequently, there's also no need to update dependency versions unless we need to use a new feature only introduced in that version -- any downstream users will automatically use whichever compatible version their binary crate wants, even if it's a newer patch release, due to the implicit caret.

Since we're not using any new features from these dependencies, is there any other reason to prevent users using an older compatible version?

hydra:
> In Cargo, the caret is the [default operator if none are specified](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), so there's no need to add an explicit caret. 

ahh ok, I missed the 'default operator' bit in the docs, thanks for pointing that out.

> Consequently, there's also no need to update dependency versions unless we need to use a new feature only introduced in that version -- any downstream users will automatically use whichever compatible version their binary crate wants, even if it's a newer patch release, due to the implicit caret.

> Since we're not using any new features from these dependencies, is there any other reason to prevent users using an older compatible version?

No, I'll close this PR then, That was easy! 😄

PR #362, opened 2021-11-12
add note

gnxlxnxx:
the same note as in https://github.com/rust-embedded/riscv/pull/87


jonathanpallant (approved):

jonathanpallant:
Hi! Thank you for this PR :) As you made a commit to your own copy of `master`, I suggest you now delete your fork - otherwise your copy will forever contain one more commit than our copy, and this may cause confusion in the future. Thanks again!

PR #363, opened 2021-11-17
scb: derive serde, Hash, PartialOrd for VectActive behind gates

tmplt:
Exposes two new feature gates for VectActive serde::{Serialize,
Deserialize} (via "serde") and Hash, PartialOrd (via "std-map") for use
on host-side ITM tracing programs. While the struct itself is not
received directly over ITM, its use greatly simplifies the
implementation by allowing VectActive as keys in map collections and
file/socket {,de}serialization to forward the structure elsewhere. These
features are not enabled by default.

Before this patch, serde functionality could be realized via [0], but
this does not propagate down a dependency chain (i.e. if realized for
crate B, which crate A depends on, serde functionality is not exposed in
crate A unless VectActive is wrapped in a type from crate B). I am not
aware of any method to realize PartialOrd, Hash derivation for a
downstream crate.

[0] https://serde.rs/remote-derive.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,11 +21,17 @@ volatile-register = "0.2.0"
 bitfield = "0.13.2"
 embedded-hal = "0.2.4"
 
+[dependencies.serde]
+version = "1"
+features = [ "derive" ]
+optional = true
+
 [features]
 cm7 = []
 cm7-r0p1 = ["cm7"]
 inline-asm = []
 linker-plugin-lto = []
+std-map = []
 
 [workspace]
 members = ["xtask", "cortex-m-semihosting", "panic-semihosting", "panic-itm"]
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -11,6 +11,8 @@ use super::CBP;
 #[cfg(not(armv6m))]
 use super::CPUID;
 use super::SCB;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 
 /// Register block
 #[repr(C)]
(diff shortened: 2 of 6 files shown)


thejpster (approved):
Looks OK to me.

thejpster:
Maybe we should expand the CI coverage to compile the crate into a non-Cortex-M host application, to check people can use the types we define for host-side purposes.

tmplt:
I'll take a look at the CI and amend fmt fixes later this week.

Cargo.toml
```diff
@@ -21,11 +21,19 @@ volatile-register = "0.2.0"
...
+features = [ "derive" ]
+optional = true
+
 [features]
 cm7 = []
 cm7-r0p1 = ["cm7"]
 inline-asm = []
 linker-plugin-lto = []
+serde = [ "serde_crate" ]
+std-map = []
```
thalesfragoso, line 36:
```suggestion
[dependencies.serde]
version = "1"
features = [ "derive" ]
optional = true

[features]
cm7 = []
cm7-r0p1 = ["cm7"]
inline-asm = []
linker-plugin-lto = []
std-map = []
```

Can't we just do this ? And get rid of the `serde(crate = "serde_crate")` on the source file ?

tmplt: I was not aware that `namespaced-features` was stabilized. Fixed.

tmplt:
Latest force-push fixes formatting and adds a `check_host_side` xtask that check serde and `PartialOrd` functionality.

thejpster (approved):

thalesfragoso (approved):
Nice work, thanks!

bors r+

PR #366, opened 2021-12-07
itm: derive serde for `LocalTimestampOptions`, impl gated `TryFrom<u8>`

tmplt:
This PR is an upstream push of more std-features required by `cargo-rtic-scope`. If required, the `TryFrom<u8>` impl can be kept downstream.


adamgreig (approved):
Thanks, seems fine to me. I've also backported renaming `std-map` to `std` for the upcoming v0.7.4 branch so it's not a surprise change from 0.7 to 0.8 (though I doubt anyone will be using what's in 0.7 anyway...).

bors merge

PR #342, opened 2021-04-28
Implement various interfaces for trace configuration

tmplt:
I'm working on tracing support and aim to implement functions that abstract the configuration of relevant peripherals. Of chief interest is `DWT`, `ITM` and `TPIU`. Some propored abstractions will go against what is established in the crate; I will ask for comments on these.

PR #367, opened 2021-12-14
add methods needed for cycle count comparisons

TDHolmes:
Extends the existing DWT comparator code to include cycle count comparisons which fire a `DebugMonitor` exception, if enabled with `enable_debug_monitor`. See working example [here](https://github.com/TDHolmes/embedded-profiling/blob/58d8b106921977816e2b5a05bcb43e976197edaf/ep-dwt/src/lib.rs#L57)

### DIFF
--- a/src/peripheral/dwt.rs
+++ b/src/peripheral/dwt.rs
@@ -387,7 +387,7 @@ impl Comparator {
     /// Configure the function of the comparator
     #[allow(clippy::missing_inline_in_public_items)]
     pub fn configure(&self, settings: ComparatorFunction) -> Result<(), DwtError> {
-        let (func, emit, data_match, cyc_match, comp, mask) = match settings {
+        match settings {
             ComparatorFunction::Address(settings) => {
                 // FUNCTION, EMITRANGE
                 // See Table C1-14
@@ -411,50 +411,45 @@ impl Comparator {
                     (_, EmitOption::PC) => return Err(DwtError::InvalidFunction),
                 };
 
-                (
-                    function,
-                    emit_range,
-                    // don't compare data value
-                    false,
-                    // don't compare cycle counter value
-                    // NOTE: only needed for comparator 0, but is SBZP.
-                    false,
-                    settings.address,
-                    settings.mask,
-                )
+                unsafe {
+                    self.function.modify(|mut r| {
+                        r.set_function(function);
+                        r.set_emitrange(emit_range);
+                        // don't compare data value
+                        r.set_datavmatch(false);
+                        // don't compare cycle counter value
+                        // NOTE: only needed for comparator 0, but is SBZP.
+                        r.set_cycmatch(false);
+
+                        r
+                    });
+
+                    self.comp.write(settings.address);
+                    self.mask.write(settings.mask);
+                }
             }
             ComparatorFunction::CycleCount(settings) => {
-                (
-                    // emit a Debug Watchpoint event, either halting execution or
-                    // firing a `DebugMonitor` exception
-                    0b0111,
-                    // emit_range is N/A for cycle count compare
-                    false,
-                    // don't compare data
-                    false,
-                    // compare cyccnt
-                    true,
-                    settings.compare,
-                    settings.mask,
-                )
+                unsafe {
+                    self.function.modify(|mut r| {
+                        // emit a Debug Watchpoint event, either halting execution or
+                        // firing a `DebugMonitor` exception
+                        // See Table C1-15 DWT cycle count comparison functions
+                        r.set_function(0b0100);
+                        // emit_range is N/A for cycle count compare
+                        r.set_emitrange(false);
+                        // don't compare data
+                        r.set_datavmatch(false);
+                        // compare cyccnt
+                        r.set_cycmatch(true);
+                        r
+                    });
+
+                    self.comp.write(settings.compare);
+                    self.mask.write(settings.mask);
+                }
             }
-        };
-
-        unsafe {
-            self.function.modify(|mut r| {
-                r.set_function(func);
-                r.set_emitrange(emit);
-                r.set_datavmatch(data_match);
-
-                // NOTE: only valid for comparator 0, but is SBZP.
-                r.set_cycmatch(cyc_match);
-
-                r
-            });
-
-            self.comp.write(comp);
-            self.mask.write(mask);
         }
+
         Ok(())
     }
 }
--- a/src/peripheral/dwt.rs
+++ b/src/peripheral/dwt.rs
@@ -392,7 +392,7 @@ pub enum ComparatorFunction {
     /// Compare cycle count & target value.
     ///
     /// **NOTE**: only supported by comparator 0 and if the HW supports the cycle counter.
-    /// Check `dwt.has_cycle_counter` for support. See C1.8.1 for more details.
+    /// Check [`DWT::has_cycle_counter`] for support. See C1.8.1 for more details.
     CycleCount(CycleCountSettings),
 }
 
--- a/src/peripheral/dwt.rs
+++ b/src/peripheral/dwt.rs
@@ -318,15 +318,15 @@ impl DWT {
 /// Whether the comparator should match on read, write or read/write operations.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 pub enum AccessType {
-    /// Generate packet only when matched adress is read from.
+    /// Generate packet only when matched address is read from.
     ReadOnly,
-    /// Generate packet only when matched adress is written to.
+    /// Generate packet only when matched address is written to.
     WriteOnly,
-    /// Generate packet when matched adress is both read from and written to.
+    /// Generate packet when matched address is both read from and written to.
     ReadWrite,
 }
 
-/// The sequence of packet(s) that should be emitted on comparator match.
+/// The sequence of packet(s) or events that should be emitted/generated on comparator match.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 pub enum EmitOption {
     /// Emit only trace data value packet.
@@ -341,6 +341,15 @@ pub enum EmitOption {
     AddressData,
     /// Emit trace PC value and data value packets.
     PCData,
+    /// Generate a watchpoint debug event.
+    ///
+    /// either halts execution or fires a `DebugMonitor` exception.
+    /// See more in section "Watchpoint debug event generation" page C1-729
+    WatchpointDebugEvent,
+    /// Generate a `CMPMATCH[N]` event.
+    ///
+    /// See more in section "CMPMATCH[N] event generation" page C1-730
+    CompareMatchEvent,
 }
 
 /// Settings for address matching
@@ -359,6 +368,9 @@ pub struct ComparatorAddressSettings {
 /// Settings for cycle count matching
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 pub struct CycleCountSettings {
+    /// The function selection used.
+    /// See Table C1-15 DWT cycle count comparison functions
+    pub emit: EmitOption,
     /// The cycle count value to compare against.
     pub compare: u32,
     /// The cycle count mask value to use.
@@ -396,16 +408,22 @@ impl Comparator {
                     (AccessType::ReadOnly, EmitOption::Address) => (0b1100, true),
                     (AccessType::ReadOnly, EmitOption::AddressData) => (0b1110, true),
                     (AccessType::ReadOnly, EmitOption::PCData) => (0b1110, false),
+                    (AccessType::ReadOnly, EmitOption::WatchpointDebugEvent) => (0b0101, false),
+                    (AccessType::ReadOnly, EmitOption::CompareMatchEvent) => (0b1001, false),
 
                     (AccessType::WriteOnly, EmitOption::Data) => (0b1101, false),
                     (AccessType::WriteOnly, EmitOption::Address) => (0b1101, true),
                     (AccessType::WriteOnly, EmitOption::AddressData) => (0b1111, true),
                     (AccessType::WriteOnly, EmitOption::PCData) => (0b1111, false),
+                    (AccessType::WriteOnly, EmitOption::WatchpointDebugEvent) => (0b0110, false),
+                    (AccessType::WriteOnly, EmitOption::CompareMatchEvent) => (0b1010, false),
 
                     (AccessType::ReadWrite, EmitOption::Data) => (0b0010, false),
                     (AccessType::ReadWrite, EmitOption::Address) => (0b0001, true),
                     (AccessType::ReadWrite, EmitOption::AddressData) => (0b0010, true),
                     (AccessType::ReadWrite, EmitOption::PCData) => (0b0011, false),
+                    (AccessType::ReadWrite, EmitOption::WatchpointDebugEvent) => (0b0111, false),
+                    (AccessType::ReadWrite, EmitOption::CompareMatchEvent) => (0b1011, false),
 
                     (AccessType::ReadWrite, EmitOption::PC) => (0b0001, false),
                     (_, EmitOption::PC) => return Err(DwtError::InvalidFunction),
@@ -429,12 +447,16 @@ impl Comparator {
                 }
             }
             ComparatorFunction::CycleCount(settings) => {
+                let function = match &settings.emit {
+                    EmitOption::PCData => 0b0001,
+                    EmitOption::WatchpointDebugEvent => 0b0100,
+                    EmitOption::CompareMatchEvent => 0b1000,
+                    _ => return Err(DwtError::InvalidFunction),
+                };
+
                 unsafe {
                     self.function.modify(|mut r| {
-                        // emit a Debug Watchpoint event, either halting execution or
-                        // firing a `DebugMonitor` exception
-                        // See Table C1-15 DWT cycle count comparison functions
-                        r.set_function(0b0100);
+                        r.set_function(function);
                         // emit_range is N/A for cycle count compare
                         r.set_emitrange(false);
                         // don't compare data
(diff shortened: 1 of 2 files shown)


TDHolmes:
Also tag @tmplt

src/peripheral/dwt.rs
```diff
@@ -400,25 +411,50 @@ impl Comparator {
...
+                    false,
+                    settings.address,
+                    settings.mask,
+                )
+            }
+            ComparatorFunction::CycleCount(settings) => {
+                (
+                    // emit a Debug Watchpoint event, either halting execution or
+                    // firing a `DebugMonitor` exception
+                    0b0111,
```
tmplt, line 430: Please comment where this function is found in reference documentation.

src/peripheral/dwt.rs
```diff
@@ -400,25 +411,50 @@ impl Comparator {
...
+                    // emit_range is N/A for cycle count compare
+                    false,
+                    // don't compare data
+                    false,
+                    // compare cyccnt
+                    true,
+                    settings.compare,
+                    settings.mask,
+                )
```
tmplt, line 439: I'm not a fan of the tuples. It'll become difficult to read when `ComparatorFunction` is further extended. Return a structure instead.

tmplt (changes requested):
Glad to see the DWT API extended. Some minor refactoring should be addressed so that the module is easily extended further in the future.

src/peripheral/dwt.rs
TDHolmes: Done. Turns out I was looking at the wrong table and this value was technically unpredictable but worked for my chip! changed to the right value and referred to the table.

src/peripheral/dwt.rs
TDHolmes: Fair point. I changed to a tuple to consolidate where we're updating the register values and cut down on duplicate code, but if we want a structure for clarity, might as well just have the duplicate locations, so I switched back to that.

src/peripheral/dwt.rs
```diff
@@ -356,12 +356,23 @@ pub struct ComparatorAddressSettings {
...
+}
+
 /// The available functions of a DWT comparator.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 #[non_exhaustive]
 pub enum ComparatorFunction {
     /// Compare accessed memory addresses.
     Address(ComparatorAddressSettings),
+    /// Compare cycle count  & target value.
+    CycleCount(CycleCountSettings),
```
tmplt, line 375: Document that this function generates a watchpoint debug event, and what section in the reference manual they can read up on it.

src/peripheral/dwt.rs
```diff
@@ -356,12 +356,23 @@ pub struct ComparatorAddressSettings {
     pub access_type: AccessType,
 }
 
+/// Settings for cycle count matching
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub struct CycleCountSettings {
```
tmplt, line 361: Could you add support for the two other cycle count comparison functions: PC value packet generation (`0b0001`) and `CMPMATCH[N]` event generation (`0b1000`)?

tmplt (changes requested):

src/peripheral/dwt.rs
tmplt: An alternative is to implement some `fn apply` on that struct.

src/peripheral/dwt.rs
tmplt: If not, please refactor in some `#[non_exhaustive]` enum that's used to select the watchpoint debug event function.

src/peripheral/dwt.rs
TDHolmes: Moved that documentation to `EmitOption`

src/peripheral/dwt.rs
TDHolmes: Added and also extended the existing code for the new `EmitOption`'s

src/peripheral/dwt.rs
```diff
@@ -318,15 +318,15 @@ impl DWT {
...
     WriteOnly,
-    /// Generate packet when matched adress is both read from and written to.
+    /// Generate packet when matched address is both read from and written to.
     ReadWrite,
 }
 
-/// The sequence of packet(s) that should be emitted on comparator match.
+/// The sequence of packet(s) or events that should be emitted/generated on comparator match.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 pub enum EmitOption {
```
tmplt, line 331: This will eventually have to be split into multiple enums, I think. Outside the scope of this PR.

src/peripheral/dwt.rs
```diff
@@ -341,6 +341,15 @@ pub enum EmitOption {
     AddressData,
     /// Emit trace PC value and data value packets.
     PCData,
+    /// Generate a watchpoint debug event.
+    ///
+    /// either halts execution or fires a `DebugMonitor` exception.
+    /// See more in section "Watchpoint debug event generation" page C1-729
```
tmplt, line 347:
```suggestion
    /// Generate a watchpoint debug event. Either halts execution or fires a `DebugMonitor` exception.
    ///
    /// See more in section "Watchpoint debug event generation" page C1-729.
```

src/peripheral/dwt.rs
```diff
@@ -341,6 +341,15 @@ pub enum EmitOption {
...
+    WatchpointDebugEvent,
+    /// Generate a `CMPMATCH[N]` event.
+    ///
+    /// See more in section "CMPMATCH[N] event generation" page C1-730
```
tmplt, line 351:
```suggestion
    /// See more in section "CMPMATCH[N] event generation" page C1-730.
```

src/peripheral/dwt.rs
```diff
@@ -356,12 +365,26 @@ pub struct ComparatorAddressSettings {
     pub access_type: AccessType,
 }
 
+/// Settings for cycle count matching
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub struct CycleCountSettings {
+    /// The function selection used.
+    /// See Table C1-15 DWT cycle count comparison functions
```
tmplt, line 372:
```suggestion
    /// See Table C1-15 for DWT cycle count comparison functions.
```

src/peripheral/dwt.rs
```diff
@@ -377,46 +400,76 @@ impl Comparator {
...
+            ComparatorFunction::Address(settings) => {
                 // FUNCTION, EMITRANGE
                 // See Table C1-14
                 let (function, emit_range) = match (&settings.access_type, &settings.emit) {
                     (AccessType::ReadOnly, EmitOption::Data) => (0b1100, false),
                     (AccessType::ReadOnly, EmitOption::Address) => (0b1100, true),
                     (AccessType::ReadOnly, EmitOption::AddressData) => (0b1110, true),
                     (AccessType::ReadOnly, EmitOption::PCData) => (0b1110, false),
+                    (AccessType::ReadOnly, EmitOption::WatchpointDebugEvent) => (0b0101, false),
+                    (AccessType::ReadOnly, EmitOption::CompareMatchEvent) => (0b1001, false),
```
tmplt, line 412: I'll review these changes later.

src/peripheral/dwt.rs
TDHolmes: Yeah, it's a bit overloaded now but so is how ARM implemented it 😅

tmplt: I started the review thinking I had previously covered all emit options for address comparisons but that was not the case. No refactor needed at the moment.

src/peripheral/dwt.rs
tmplt: LGTM

src/peripheral/dwt.rs
```diff
@@ -356,12 +364,26 @@ pub struct ComparatorAddressSettings {
...
+}
+
 /// The available functions of a DWT comparator.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 #[non_exhaustive]
 pub enum ComparatorFunction {
     /// Compare accessed memory addresses.
     Address(ComparatorAddressSettings),
+    /// Compare cycle count  & target value.
+    CycleCount(CycleCountSettings),
```
tmplt, line 386:
```suggestion
    /// Compare cycle count  & target value.
    ///
    /// **NOTE**: only supported by comparator 0. See C1.8.1.
    CycleCount(CycleCountSettings),
```

tmplt: Ideally we should statically check what comparator is being configured, but I don't know whether that is possible.

src/peripheral/dwt.rs
```diff
@@ -356,12 +364,26 @@ pub struct ComparatorAddressSettings {
...
+/// Settings for cycle count matching
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub struct CycleCountSettings {
+    /// The function selection used.
+    /// See Table C1-15 for DWT cycle count comparison functions.
+    pub emit: EmitOption,
+    /// The cycle count value to compare against.
+    pub compare: u32,
+    /// The cycle count mask value to use.
+    pub mask: u32,
```
tmplt, line 376: Remove this: `DWT_MASK` is SBZ on cycle count comparison.

src/peripheral/dwt.rs
```diff
@@ -377,46 +399,76 @@ impl Comparator {
...
+                        r.set_emitrange(false);
+                        // don't compare data
+                        r.set_datavmatch(false);
+                        // compare cyccnt
+                        r.set_cycmatch(true);
+                        r
+                    });
+
+                    self.comp.write(settings.compare);
+                    self.mask.write(settings.mask);
```
tmplt, line 469:
```suggestion
                    self.mask.write(0); // SBZ, see Page 784
```

src/peripheral/dwt.rs
```diff
@@ -377,46 +399,76 @@ impl Comparator {
...
+                        r.set_cycmatch(false);
+
+                        r
+                    });
+
+                    self.comp.write(settings.address);
+                    self.mask.write(settings.mask);
+                }
+            }
+            ComparatorFunction::CycleCount(settings) => {
```
tmplt, line 448: Add a check to ensure that `DWT_CTRL.NOCYCCNT` is RAZ. Otherwise, return `Err(DwtError::NoCycCnt)`.

tmplt (changes requested):
Other than the above, `DATAV{ADDR{1,0},SIZE}` are SBZ on cycle count match, but these are not declared in the `Function` bitfield (c.f. Page 806) and never written to. These fields are also SBZ on address comparison match. I wonder if it isn't a good idea to write zero to these just in case.

src/peripheral/dwt.rs
TDHolmes: I couldn't think of a way to do so without a more invasive change, like special casing the first two as special comparators. I'm not convinced it'll be really worth it..

tmplt:
> These fields are also SBZ on address comparison match. I wonder if it isn't a good idea to write zero to these just in case.

It's for the better now that I think about it. If they are non-zero target behavior will be unpredictable on comparator match. Had they been marked SBZP I would have disregarded it, but that's not the case.

src/peripheral/dwt.rs
TDHolmes: Added the note

TDHolmes:
Agreed, updated

src/peripheral/dwt.rs
```diff
@@ -82,11 +82,17 @@ bitfield! {
...
     /// Comparator FUNCTIONn register.
+    ///
+    /// See C1.8.17 "Comparator Function registers, DWT_FUNCTIONn"
     pub struct Function(u32);
     u8, function, set_function: 3, 0;
     emitrange, set_emitrange: 5;
     cycmatch, set_cycmatch: 7;
     datavmatch, set_datavmatch: 8;
+    lnk1ena, set_lnk1ena: 9;
+    u8, datavsize, set_datavsize: 2, 10;
```
tmplt, line 93:
```suggestion
    u8, datavsize, set_datavsize: 11, 10;
```

src/peripheral/dwt.rs
```diff
@@ -82,11 +82,17 @@ bitfield! {
...
+    u8, datavaddr0, set_datavaddr0: 4, 12;
```
tmplt, line 94:
```suggestion
    u8, datavaddr0, set_datavaddr0: 15, 12;
```

src/peripheral/dwt.rs
```diff
@@ -82,11 +82,17 @@ bitfield! {
...
+    u8, datavaddr1, set_datavaddr1: 4, 16;
```
tmplt, line 95:
```suggestion
    u8, datavaddr1, set_datavaddr1: 19, 16;
```

src/peripheral/dwt.rs
```diff
@@ -370,53 +401,94 @@ pub enum ComparatorFunction {
...
     InvalidFunction,
+    /// The DWT block does not implement cycle count capabilities.
+    NoCycleCount,
 }
 
 impl Comparator {
     /// Configure the function of the comparator
     #[allow(clippy::missing_inline_in_public_items)]
-    pub fn configure(&self, settings: ComparatorFunction) -> Result<(), DwtError> {
+    pub fn configure(&self, dwt: &DWT, settings: ComparatorFunction) -> Result<(), DwtError> {
```
tmplt, line 411: I misread the standard. I thought `self` could tell us if this comparator supported CYCCNT, but that information is indeed only in `DWT_NOCYCCNT`. Standard says
> Comparator 0 supports cycle count comparisons only if the DWT_CTRL.NOCYCCNT bit is RAZ

but we have no way to check if this is comparator zero; with the new check CYCCNT can still be configured for a comparator that isn't zero.

We'll have to compromise: revert the new check and add a note in the documentation.

src/peripheral/dwt.rs
```diff
@@ -370,53 +401,94 @@ pub enum ComparatorFunction {
...
-                });
+                unsafe {
+                    self.function.modify(|mut r| {
+                        r.set_function(function);
+                        r.set_emitrange(emit_range);
+                        // don't compare data value
+                        r.set_datavmatch(false);
+                        // don't compare cycle counter value
+                        // NOTE: only needed for comparator 0, but is SBZP.
+                        r.set_cycmatch(false);
```
tmplt, line 450:
```suggestion
                        r.set_cycmatch(false);
                        // SBZ as needed, see Page 784/C1-724
                        r.set_datavsize(0);
                        r.set_datavaddr0(0);
                        r.set_datavaddr1(0);
```

tmplt (changes requested):
This should be the last batch.

src/peripheral/dwt.rs
TDHolmes: I don't think it's the worst thing in the world to **also** check if the HW is even capable of `CYCCNT`. The documentation for `ComparatorFunction::CycleCount` already mentions the 0th comparator requirement

tmplt: The root issue in my eyes is requiring the user to pass the `DWT` to `configure` for an effectivly non-exhaustive check: while the current check ensures that `dwt[0]` supports cycle counters, `configure` still allows us to configure `dwt[1]` for it (effectively an invalid `configure` invocation, hardware-wise), as long as `dwt[0]` supports it. An exhaustive check would require two different types for `dwt[0]` and `dwt[n > 0]` and only allow cycle count comparisons for the `dwt[0]` type (as previously discussed), but that's an issue for another day.

TDHolmes:
Yeah fair argument. I’ll remove the check and add to the documentation a
recommendation for the end user to check this function if they’re going to
use cycle compare

TDHolmes:
CI is failing due to #371

src/peripheral/dwt.rs
```diff
@@ -356,12 +373,27 @@ pub struct ComparatorAddressSettings {
...
 /// The available functions of a DWT comparator.
 #[derive(Debug, Eq, PartialEq, Copy, Clone)]
 #[non_exhaustive]
 pub enum ComparatorFunction {
     /// Compare accessed memory addresses.
     Address(ComparatorAddressSettings),
+    /// Compare cycle count & target value.
+    ///
+    /// **NOTE**: only supported by comparator 0 and if the HW supports the cycle counter.
+    /// Check `dwt.has_cycle_counter` for support. See C1.8.1 for more details.
```
tmplt, line 395:
```suggestion
    /// Check [DWT::has_cycle_counter] for support. See C1.8.1 for more details.
```

tmplt (changes requested):
Thanks for your work.

src/peripheral/dwt.rs
TDHolmes: I had initially tried this but rust-analyzer didn't highlight it as valid, however checking `cargo doc` looks like it can. Fixed, but using "`"'s for fixed-width text formatting.

tmplt (approved):

src/peripheral/dwt.rs
tmplt: It is a valid link then? Mustn't you use the below?

    /// [`DWT::has_cycle_counter`](DWT::has_cycle_counter)

TDHolmes: [`DWT::has_cycle_compare`] wasn't highlighting for some reason so I had dropped the square brackets. Seems like just a bug in the highlighter or rust-analyzer 🤷

TDHolmes:
@tmplt thanks for the groundwork and thorough review! Looking forward to reading your thesis 😄

thejpster:
Are those nightly CI failures expected due to asm! changes?

TDHolmes:
Yes

TDHolmes:

TDHolmes:
Any update on this PR? It should be a non breaking change. Even though `EmitOption` is public and has new variants, it's only used in constructing configuration structs so it being not marked as `nonexhaustive` shouldn't be a problem

thejpster:
Looks ok to me, but I'm not an expert in this area

TDHolmes:
@tmplt Seems to be the expert and approved but I dunno if he has privileges to merge

adamgreig (approved):
This is definitely a breaking change as it adds variants to two public enums that aren't already `non_exhaustive`, but all the code it changes is already breaking changes from #342, so it will have to go into a 0.8 either way.

Other than that LGTM.

bors merge

PR #369, opened 2021-12-15
Switch "native" check from being x86_64 only to checking `HOST`

TDHolmes:
If `HOST==TARGET`, we know we're compiling natively. Set a new `rustc` cfg for this and use it where we previously checked for `x86_64`.

I was trying to run tests on my M1 MacBook Pro and couldn't since it isn't `x86_64`. Also, the currently configured nightly compiler for asm doesn't have M1 support, so I updated that. I'm fine reverting that change though, I can just do that locally, but I'm sure others will hit the same issue and it's a bit old...

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,8 +3,11 @@ use std::{env, fs};
 
 fn main() {
     let target = env::var("TARGET").unwrap();
+    let host_triple = env::var("HOST").unwrap();
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
     let name = env::var("CARGO_PKG_NAME").unwrap();
+
+    let native_compilation = host_triple == target;
 
     if target.starts_with("thumb") {
         let suffix = if env::var_os("CARGO_FEATURE_LINKER_PLUGIN_LTO").is_some() {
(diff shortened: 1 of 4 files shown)


TDHolmes:
🏷️ @japaric since it looks like rust-highfive doesn't like this PR and you modified this area of the build script previously

build.rs
```diff
@@ -43,7 +46,7 @@ fn main() {
         println!("cargo:rustc-cfg=armv8m_main");
     }
 
-    if target.ends_with("-eabihf") {
-        println!("cargo:rustc-cfg=has_fpu");
+    if native_compilation {
+        println!("cargo:rustc-cfg=native");
```
adamgreig, line 50: Is removing the `has_fpu` setting deliberate?

TDHolmes: whoops, this is definitely wrong. Thanks

TDHolmes:
Failing until #372 merges

TDHolmes:

TDHolmes:
This is pretty straight forward and enables ARM computers to run the CI

adamgreig (approved):
Looks good to me, thanks!

bors merge

PR #373, opened 2021-12-17
`SCB.ICSR.VECTACTIVE` is 9 bits, not 8

### LINKED ISSUE #332: SCB::vect_active() should be 9 bits
Currently `vect_active()` returns the irq number as a u8, but the SCB.ICSR.VECTACTIVE field is 9 (or more?) bits. This means that the interrupt numbers will wrap on chips with a lot of interrupts.

TDHolmes:
Closes #332

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -170,9 +170,10 @@ impl SCB {
     /// Returns the active exception number
     #[inline]
     pub fn vect_active() -> VectActive {
-        let icsr = unsafe { ptr::read(&(*SCB::ptr()).icsr as *const _ as *const u32) };
+        let icsr =
+            unsafe { ptr::read_volatile(&(*SCB::ptr()).icsr as *const _ as *const u32) } & 0x1FF;
 
-        match icsr as u8 {
+        match icsr as u16 {
             0 => VectActive::ThreadMode,
             2 => VectActive::Exception(Exception::NonMaskableInt),
             3 => VectActive::Exception(Exception::HardFault),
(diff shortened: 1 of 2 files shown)


src/peripheral/scb.rs
```diff
@@ -300,7 +301,7 @@ impl VectActive {
             12 => VectActive::Exception(Exception::DebugMonitor),
             14 => VectActive::Exception(Exception::PendSV),
             15 => VectActive::Exception(Exception::SysTick),
-            irqn if irqn >= 16 => VectActive::Interrupt { irqn },
+            irqn if irqn >= 16 && irqn < 512 => VectActive::Interrupt { irqn: irqn - 16 },
```
TDHolmes, line 304: the `- 16` was added to be consistent with how `vect_active` does it

src/peripheral/scb.rs
```diff
@@ -170,9 +170,10 @@ impl SCB {
     /// Returns the active exception number
     #[inline]
     pub fn vect_active() -> VectActive {
-        let icsr = unsafe { ptr::read(&(*SCB::ptr()).icsr as *const _ as *const u32) };
+        let icsr =
+            unsafe { ptr::read_volatile(&(*SCB::ptr()).icsr as *const _ as *const u32) } & 0x1FF;
```
TDHolmes, line 174: changed to `read_volatile` like `cortex-m-rt` did to prevent optimizations here

src/peripheral/scb.rs
thejpster: Is that a breaking change?

TDHolmes: Yes, but I think the rest of this PR is too

TDHolmes:

adamgreig:
Thanks! I think we can merge this now that master is explicitly for 0.8 development, but could you add a changelog entry, especially highlighting the `- 16` change to irq number?

TDHolmes:
whoops, sure! I also added an entry for my change in #367 since I forgot that too.

adamgreig (approved):
Thanks!

bors merge

PR #374, opened 2021-12-29
Construct a SysTick delay with a clock source

mciantyre:
Helpful for users who prefer an external clock source for their system
timer. Proposing this as a 0.7 backwards-compatible change.

PR #351, opened 2021-08-30
Replace URL-relative doc links with intra-doc links

newAM:
This fixes invalid documentation links for crates that re-export `cortex-m`.

PR #349, opened 2021-08-08
Add support for additional DWT counters

newAM:
This adds support for these DWT counters:
* CPI counter
* Exception overhead counter
* LSU counter
* Folded-instruction counter

The getters do not have the `get_` prefix per the [C-GETTER](https://rust-lang.github.io/api-guidelines/naming.html#c-getter) API guidelines. This does not match the existing `get_cycle_count` method. I was not sure if I should maintain consistency or not, let me know if you want this changed.

Here is the documentation to validate each of the added methods:
* [ARMv7-M DWT register summary](https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture/ARMv7-M-Debug/The-Data-Watchpoint-and-Trace-unit/DWT-register-summary)
    * [Control Register](https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture/ARMv7-M-Debug/The-Data-Watchpoint-and-Trace-unit/Control-register--DWT-CTRL)
        * `num_comp`
            * [ARMv6-M DWT register summary](https://developer.arm.com/documentation/ddi0419/c/Debug-Architecture/ARMv6-M-Debug/The-Data-Watchpoint-and-Trace-unit/DWT-register-summary)
            * The `num_comp` method is not guarded with `#[cfg(not(armv6m))]` because it is valid for both.
        * `has_exception_trace`
        * `has_external_match`
        * `has_cycle_counter`
        * `has_profiling_counter`
        * `disable_cycle_counter`
        * `cycle_counter_enabled`
    * [CPI count register](https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture/ARMv7-M-Debug/The-Data-Watchpoint-and-Trace-unit/CPI-Count-register--DWT-CPICNT)
        * `cpi_count`
        * `set_cpi_count`
    * [Exception Overhead Count register](https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture/ARMv7-M-Debug/The-Data-Watchpoint-and-Trace-unit/Exception-Overhead-Count-register--DWT-EXCCNT)
        * `exception_count`
        * `set_exception_count`
    * [Sleep Count register](https://developer.arm.com/documentation/ddi0403/d/Debug-Architecture/ARMv7-M-Debug/The-Data-Watchpoint-and-Trace-unit/Sleep-Count-register--DWT-SLEEPCNT)
        * `sleep_count`
        * `set_sleep_count`
[...]

PR #347, opened 2021-08-06
Add DWT.set_cycle_count

newAM:
This method is helpful to reset the cycle count when using it as a timestamp.

PR #375, opened 2021-12-31
Prepare v0.7.4

adamgreig:
I've created a new branch, `v0.7.x`, which is currently at the latest non-breaking commit (so includes #346 #349 #347 #351 #339 #352 #348 #363 #362 #361 but does not include #342), to track the 0.7 series since master now contains breaking changes for v0.8.

This PR (which targets the new branch) cherry-picks #372 #369 #374 and bumps the version to v0.7.4 (and updates CHANGELOG) ready for a new v0.7.4 release. Once complete I'll also backport the changelog entries and bump the version in master to 0.7.4.

I think this is everything that should be in 0.7 -- the only excluded PRs from master are #342 and #367 I believe, and I don't think we have any open PRs targeting 0.7 either.

Any other thoughts on items for inclusion in 0.7.4 (or other changelog entries I missed)?


TDHolmes:
LGTM, but wanted to ask, the bin files are much smaller now. Did we notice that back when the inline asm changes were originally updated? I see you said:
> With this change the disassembly output is identical to what we have now (phew).

so are the file size changes all just like debug info?

adamgreig:
Good spot, that is kind of alarming. I'll check.

adamgreig:
I think this is mostly a quirk of GitHub's UI - the percentage in red is the new size relative to the old size, not the size of the delta, so `-1.73 KB (89%)` means the new 14kB file is 1.73kB smaller than the old 16kB file, but we haven't lost 89% of the contents. Checking the archive files manually the difference does seem to be mostly in some tweaked debug info and various other apparently minor changes due to the new LLVM version. Overall doesn't look like a problem.

thejpster (approved):
LGTM

adamgreig:
OK, let's get this out there.

bors r=thejpster

Issue #183, opened 2020-01-07
Ship a 1.0 release


thejpster:
Part of https://github.com/rust-embedded/wg/issues/383 

Might also involve https://github.com/rust-lang/rust/issues/27731 so we can get rid of (some) of the inline asm / pre-compiled objects.

jonas-schievink:
> Might also involve [rust-lang/rust#27731](https://github.com/rust-lang/rust/issues/27731) so we can get rid of (some) of the inline asm / pre-compiled objects.

This is not needed for 1.0 since it doesn't change the public API

jonas-schievink:
Easy (breaking) improvements to do before 1.0:
* Remove the `const-fn` feature
* Get https://github.com/rust-embedded/cortex-m/pull/181 landed and audit the crate for other things that should be `#[cfg]`d out depending on the target

Other 1.0-quality things that I'd like to see:
* Improve overall documentation
* Document that the `inline-asm` feature is unstable (semver-exempt) and nighly-only

This is also blocked on sound MMIO. Currently we use `volatile-register` here. As #184 mentions, there's also a public dependency on `aligned` that we should get rid of (or also bring to 1.0).

adamgreig:
I'd really want to see cortex-m move to using svd2rust-generated API for accessing the Cortex-M peripherals, i.e. a cortex-m-pac crate that cortex-m depends on and adds the HAL API. I've made some progress towards generating a suitable set of SVD files already and plan to finish it fairly soon.

thejpster:
> 
> 
> I'd really want to see cortex-m move to using svd2rust-generated API for accessing the Cortex-M peripherals, i.e. a cortex-m-pac crate that cortex-m depends on and adds the HAL API. I've made some progress towards generating a suitable set of SVD files already and plan to finish it fairly soon.

Yeah, this would likely fundamentally change our public API so if we are ever going to do it, pre 1.0 is best.

jonas-schievink:
Also:
* Remove currently `#[deprecated]` unsound APIs (I expect we'll publish 0.7 before 1.0, so it would be good to do all breaking changes there)

adamgreig:
I'd add:

* Check all peripheral methods that could be static are, i.e. don't unnecessarily require a reference to the peripheral for safe/atomic operations. We've had a number of breaking changes moving things from methods to associated functions and the latter are easier to use, so we should proactively see if we can do any more.

Disasm:
[...]

PR #377, opened 2022-01-02
Differentiate comparator 0 as the only one capable of cycle compare

### LINKED ISSUE #376: DWT: differentiate the first comparator
As discussed in #364 we should ideally check whether a DWT feature is supported statically. This is only realistically applicable for cycle count comparisons (but still requires a runtime check) which is only supported on the first comparator. An implementation would then have some `DWTComparatorFirst` and `DWTComparator` structures that implement a common trait: if cycle count comparison is configured for `DWTComparator`, an error is returned.

Opening this issue for discussion purposes for now. At present the documentation warns that cycle count comparison is only available with the first comparator.

TDHolmes:
Statically enforces that only comparator 0 on `armv7m` can be configured for cycle comparison by introducing a marker trait differentiating comparator 0 and the rest of them, and only implementing the ability for this configuration on comparator 0.

Closes #376

### DIFF
--- a/src/peripheral/dwt.rs
+++ b/src/peripheral/dwt.rs
@@ -37,10 +37,13 @@ pub struct RegisterBlock {
     pub pcsr: RO<u32>,
     /// Comparators
     #[cfg(armv6m)]
-    pub c: [Comparator; 2],
+    pub comp: [Comparator<NoCycleCompare>; 2],
+    #[cfg(not(armv6m))]
+    /// Cycle count compare enabled Comparator
+    pub comp0: Comparator<HasCycleCompare>,
     #[cfg(not(armv6m))]
     /// Comparators
-    pub c: [Comparator; 16],
+    pub comp: [Comparator<NoCycleCompare>; 15],
     #[cfg(not(armv6m))]
     reserved: [u32; 932],
     /// Lock Access
@@ -66,9 +69,31 @@ bitfield! {
     u8, numcomp, _: 31, 28;
 }
 
+mod private {
+    /// A public trait inaccessible by external users to ensure no one else can
+    /// impl a sealed trait outside of the crate of origin. For more info on this
+    /// design pattern, see https://rust-lang.github.io/api-guidelines/future-proofing.html
+    pub trait Sealed {}
+}
+
+/// A zero-sized marker trait indicating the capabilities of a given comparator.
+pub trait ComparatorSupportedFunctions: private::Sealed {}
+
+/// Marker indicating that this comparator has cycle comparison abilities. This
+/// is the case only for the first comparator for armv7m.
+pub enum HasCycleCompare {}
+impl ComparatorSupportedFunctions for HasCycleCompare {}
+impl private::Sealed for HasCycleCompare {}
+
+/// Marker indicating this comparator does not have cycle comparison abilities. This
+/// is the case for all armv6m comparators and comparators 1-15 for armv7m.
+pub enum NoCycleCompare {}
+impl ComparatorSupportedFunctions for NoCycleCompare {}
+impl private::Sealed for NoCycleCompare {}
+
 /// Comparator
 #[repr(C)]
-pub struct Comparator {
+pub struct Comparator<SupportedFunctions: ComparatorSupportedFunctions> {
     /// Comparator
     pub comp: RW<u32>,
     /// Comparator Mask
@@ -76,6 +101,7 @@ pub struct Comparator {
     /// Comparator Function
     pub function: RW<Function>,
     reserved: u32,
+    _supported_functions: core::marker::PhantomData<SupportedFunctions>,
 }
 
 bitfield! {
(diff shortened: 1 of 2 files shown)


TDHolmes:
🏷️ @tmplt

TDHolmes:
I tested the cycle count comparison functionality on HW to make sure this all worked as expected. I'm not sure how to test the address compare functionality on HW so didn't check that.

tmplt:
I don't see any outstanding issues reviewing from a phone at the moment. I'll go over it in detail when I'm back in my office next week.

TDHolmes:
Thoughts @tmplt ?

tmplt:
I'll go over it tomorrow.

src/peripheral/test.rs
```diff
@@ -42,12 +42,21 @@ fn dwt() {
...
+
+        assert_eq!(address(&dwt.comp[0].comp), 0xE000_1030);
+        assert_eq!(address(&dwt.comp[0].mask), 0xE000_1034);
+        assert_eq!(address(&dwt.comp[0].function), 0xE000_1038);
+    }
+    if cfg!(armv6m) {
+        assert_eq!(address(&dwt.comp[0].comp), 0xE000_1020);
+        assert_eq!(address(&dwt.comp[0].mask), 0xE000_1024);
+        assert_eq!(address(&dwt.comp[0].function), 0xE000_1028);
+    }
```
tmplt, line 58: Can we test `configure` here?

TDHolmes: I don't think so unfortunately. These tests run on the native host so the best we can do is make sure these things are pointing to the right address. If we dereferenced them and tried to check behavior, we'd crash and burn

src/peripheral/dwt.rs
```diff
@@ -66,16 +69,39 @@ bitfield! {
...
-pub struct Comparator {
+pub struct Comparator<SupportedFunctions: ComparatorSupportedFunctions> {
     /// Comparator
     pub comp: RW<u32>,
     /// Comparator Mask
     pub mask: RW<u32>,
     /// Comparator Function
     pub function: RW<Function>,
     reserved: u32,
+    _supported_functions: core::marker::PhantomData<SupportedFunctions>,
```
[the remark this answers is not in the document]
TDHolmes: I'm not 100% sure what you mean so I'll take a stab at what I think you're asking. We need to "cary" around the generic argument `SupportedFunctions`, but it's zero sized so we use `PhantomData`. As the tests show, it does not increase the size of the `RegisterBlock` (addresses for `comp0` and `comp[0]` remained the same)

TDHolmes:
I made all of the `DWT::has_*` implementation feature checks associated functions instead of methods (they no longer take `&self`) per [this comment](https://github.com/rust-embedded/cortex-m/issues/183#issuecomment-579430492) as it allows me to check `has_cycle_counter` in `configure`.

TDHolmes:
I also tried to be fancy by wrapping these two slightly different comparators in a unified struct which then ventured the right one by implementing the `Index` trait, however it requires GATs to work :( it would've been cool because then we could've also checked `DWT::num_comp` to see if the comparator you're trying to access is even implemented. Oh well...

TDHolmes:
Actually.. I suppose I could still do an iterator where I return a type `Result<ComparatorTypes, DwtError>` where `ComparatorTypes` is an enum:
```rust
enum ComparatorTypes {
    NoCycleCount(Comparator<NoCycleCompare>),
    WithCycleCount(Comparator<HasCycleCompare>),
```

thoughts?

tmplt:
> Let me know if there was a good reason

At the time I thought it was relics from an older implementation which I refactored just to remove the `unsafe` usage.

As for `Index` I suggest limiting this PR to its namesake and create another one for the comparator implementation check.

TDHolmes:
In that case this should be ready to review and merge

src/peripheral/dwt.rs
```diff
@@ -101,29 +127,33 @@ impl DWT {
...
     }
 
     /// Returns `true` if the implementation supports a cycle counter
     #[inline]
-    pub fn has_cycle_counter(&self) -> bool {
+    pub fn has_cycle_counter() -> bool {
         #[cfg(not(armv6m))]
-        return !self.ctrl.read().nocyccnt();
+        // NOTE(unsafe) atomic read with no side effects
+        return !unsafe { (*DWT::PTR).ctrl.read().nocyccnt() };
```
tmplt, line 156:
```suggestion
        return unsafe { !(*DWT::PTR).ctrl.read().nocyccnt() };
```

CHANGELOG.md
```diff
@@ -20,6 +25,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Fixed `singleton!()` statics sometimes ending up in `.data` instead of `.bss` (#364, #380).
+- Corrected `SCB.ICSR.VECTACTIVE`/`SCB::vect_active()` to be 9 bits instead of 8.
+  Also fixes `VectActive::from` to take a `u16` and subtract `16` for
+  `VectActive::Interrupt`s to match `SBC::vect_active()` (#373).
```
tmplt, line 30: Transient changes from another PR?

src/peripheral/dwt.rs
```diff
@@ -132,8 +162,9 @@ impl DWT {
     /// Returns `true` if the implementation the profiling counters
     #[cfg(not(armv6m))]
     #[inline]
-    pub fn has_profiling_counter(&self) -> bool {
-        !self.ctrl.read().noprfcnt()
+    pub fn has_profiling_counter() -> bool {
+        // NOTE(unsafe) atomic read with no side effects
+        !unsafe { (*DWT::PTR).ctrl.read().noprfcnt() }
```
tmplt, line 167:
```suggestion
        unsafe { !(*DWT::PTR).ctrl.read().noprfcnt() }
```

tmplt (changes requested):

tmplt:
I want to try this out on hardware before I sign off on it. Hopefully I'll have it done by next week.

TDHolmes:
Sounds good, thanks!

tmplt:
I've been testing this PR with a in `cortex-m-rtic-trace` branch, based on an `rtic-scope-dw0` branch of cortex-m that contains this PR, for use with `cargo-rtic-scope`. Without the patch I see 
```
hardware task - enter
software task - enter
software task - exit
hardware task - exit / overflow
```
With the patch I get a very confusing trace. I doubt this PR is at fault, but to make sure I need to handle #392 and #383 first.

TDHolmes:
rebased onto latest master

PR #380, opened 2022-01-03
Improve singleton!() macro

### LINKED ISSUE #364: singleton!()'s use of Option can have nasty optimization results
I was trying to store some large buffers inside of a `cortex_m::singleton!()` when I noticed that suddenly my flash usage increased dramatically.  As some (contrived) examples, the following works (variable ends up in `.bss`):

```rust
cortex_m::singleton!(: [u8; 2048] = [0xff; 2048]);
```

but this doesn't (variable ends up in `.data`):

```rust
cortex_m::singleton!(: [bool; 2048] = [true; 2048]);
```

The reason is the `Option` inside the `singleton!()` macro:

https://github.com/rust-embedded/cortex-m/blob/6b013138b734b9bbeb24a345f75d2bcc1c69fa8d/src/macros.rs#L55

Normally, the `None` value which is stored here has a bitpattern of all zeros, leading to it being stored in `.bss`.  But when the given type has a niche (like `bool` above does), the `None` variant is "optimized" into it.  This means it then has non-zero bits and cannot be stored in `.bss` anymore, leading to it getting moved to `.data`...

A quick solution is to "reimplement" the `Option`:

```rust
	static mut VAR: (MaybeUninit<$ty>, bool) = (MaybeUninit::uninit(), false);
```

This will always end up in `.bss` because no niche optimization exists for such types (yet).  Before opening a PR, I wanted to discuss whether there are any objections to changing to such an implementation?

Side Note: I think it would be nice to allow users to overwrite the name of the `static`.  When looking at `nm` or `bloaty` output, only seeing a bunch of symbols named `__cortex_m_rt_main::{{closure}}::VAR` is not great...

Rahix:
This PR addresses two shortcomings of the `cortex_m::singleton!()` macro, which I raised in #364.  For review, I think it is best to look at the two commits implementing these changes individually.

I think this changeset should also be backported to `0.7.x` where it applies cleanly and which is also the place where I tested it.

### 1. Fix `singleton!()` sometimes ending up in `.data`
The static is always initialized to a "zero" value with `Option::None` which means it should end up in `.bss`.  However, if the enclosed type has a niche, `Option::None` can become a non-zero bitpattern which moves the whole singleton from `.bss` to `.data`.  This is especially problematic when storing large buffers in the `singleton!()` as this starts eating lots of flash space unnecessarily.

To prevent this, I switched to using an explicit boolean flag instead.  This is not quite as nice but at least there is no chance for the `singleton!()` to end up in `.data`...

For reference and as an example, the faulty behavior can be triggered with

```rust
cortex_m::singleton!(: Option<u32> = None)
```

(the inner option has a non-zero niche which the outer option will occupy)

### 2. Allow naming the `static`
Due to the static always being named `VAR` right now, all `singleton!()` instances end up having non-descriptive symbol names like `__cortex_m_rt_main::{{closure}}::VAR` which makes them hard to tell apart in a debugger or when looking at an objdump.

I added the ability to set an explicit name which end up becoming part of the symbol name.  This does not affect Rust code at all - the new name is not visible anywhere.  It works like this:

```rust
let value = singleton!(FOO_BUFFER: [u8; 1024] = [0u8; 1024]);
```

Of course the old syntax also still works and keeps the old behavior of calling the static `VAR`.

Fixes #364.


TDHolmes:
Can you add a CHANGELOG entry?

thejpster (approved):
Also looks good to me - just the needs the changelog update as noted.

Rahix:
Added CHANGELOG entries.  Should I send a second PR against the `v0.7.x` branch or will someone else take care of backports?

adamgreig:
Thanks, this looks good to me too. Would you mind adding a comment inline to the macro definition briefly mentioning why it's a tuple with a bool instead of an Option, to help anyone coming along in the future with a great idea for an optimisation?

I can take care of the backport once this is merged. I'll probably give it a little while though in case anything else comes along that can be included in a new 0.7 release since we only just pushed 0.7.4 a couple of days ago.

Rahix:
@adamgreig: Good idea, done.

Issue #382, opened 2022-01-04
Verify peripheral configuration support inline instead of seperate `has_` functions?


tmplt:
Throughout the crate I believe most (if not all) register-mutating functions are unchecked, meaning that the crate user should check for support before enabling and configurating a feature. I consider this reasonable for trivial functions such as `DWT::has_exception_trace` and `DWT::enable_exception_tracing` but less so for more complex functions like `DWT::configure` and `ITM::configure` which may enable and configure a whole set of features. For these functions it's not trivial to figure out which `has_` functions must be called before  configuration (presuming they are implemented in the first place).

I believe it would be a better idea to verify support in `configure` functions before registers are mutated (and return `Err` if the requested configuration cannot be applied) and offer some `configure_unchecked` variant when the user knows when a set of features are supported. I'm not sure if we should do the same to the more drivial functions.

Thoughts?

tmplt:
An edge-case to consider are RAZ/WI (Read-as-zero, writes ignored) register fields which do not necessarily have an associated flag bit to check for support. Per the standard one should try to configure these and if they subsequenty read zero (unless zero was written to disable the feature), the configuration failed. An example of this is `ITM_TCR.GTSFREQ`. Executing an `ITM::has_global_timestamps` could lead to unintended side-effects because one must:
1. read the flag value and store it;
2. write something non-zero to the flag location (all non-zero values having side-effects)
3. read the flag value again;
4. compare the old and new flag values;
5. if flags are equal (and both thus zero), return false; or
6. if flags differ, write the old value back and return true.

A more proper way to achieve this is to read back the `GTSFREQ` value in `ITM::configure` and return failure if the user requested global timestamp generation and it wasn't supported. This must be done before any other bits are flipped, so that we don't end up with a partially applied configuration. In cases where a potentially RAZ/WI field cannot be touched before another field is configured, setup should be split this into some unlock function (e.g. `ITM::unlock`) when it makes sense or document it. Fortunately I have yet to see any heavy nesting of this sort.

tmplt:
See above PR. Implementing an `ITM::configure_unchecked` would in this case lead to a lot of code duplication due to the lack of feature support flags.

TDHolmes:
I don't think the code duplication would be too bad. Just pulling out all of the write logic to `*_unchecked` and use that function in the safe version after doing all of the checking. You'd do a bit of double checking of the `settings` struct but it's probably not too bad and probably inlines away. 

[...]

PR #383, opened 2022-01-04
ITM: check feature support during configuration

tmplt:
This PR is a test realization of #382.


tmplt:
dc4642512761f2b6d9fe65d7878cd12a2131aeb8 can be fixup'd into 622a9b7297f47a1d6a0d40894e2d2994ad46db27.

src/peripheral/itm.rs
```diff
@@ -174,42 +174,119 @@ pub struct ITMSettings {
     pub timestamp_clk_src: TimestampClkSrc,
 }
 
+/// Possible errors on [ITM::configure].
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub enum ITMConfigurationError {
+    /// Global timestamp generation is not supported on this target.
+    /// Request [GlobalTimestampOptions::Disabled] instead.
+    ///
+    /// `ITM_TCR` register remains unchanged on this error.
```
TDHolmes, line 183: I think you can do the following to make it an intra-doc link:
```suggestion
    /// [`ITM_TCR`](struct@Tcr) register remains unchanged on this error.
```
And this can be extrapolated to the rest. I think it's also best to make the inner mono spaced with \`'s

src/peripheral/itm.rs
```diff
@@ -174,42 +174,119 @@ pub struct ITMSettings {
...
+    /// [LocalTimestampOptions::Disabled] instead.
+    ///
+    /// *NOTE*: `ITM_TCR.{GTSFREQ,SWOENA}` fields have potentially
+    /// changed on this error.
+    TSPrescale,
+}
+
 impl ITM {
-    /// Removes the software lock on the ITM.
+    /// Removes the software lock on the ITM. Must be called before any other [ITM] functions.
```
TDHolmes, line 200: Might as well make all mentions of `ITM` be `[`ITM`]`

tmplt:
Wait with eventual merge for me to test these checks. It should be done by next week.

tmplt:
https://github.com/rust-embedded/cortex-m/issues/392 must be handled in this PR.

src/peripheral/itm.rs
```diff
@@ -174,42 +186,176 @@ pub struct ITMSettings {
...
+    ///
+    /// See (coresight, B2.3.10).
     #[inline]
     pub fn unlock(&mut self) {
         // NOTE(unsafe) atomic write to a stateless, write-only register
         unsafe { self.lar.write(0xC5AC_CE55) }
     }
 
-    /// Configures the ITM with the passed [ITMSettings].
+    /// Adds the software lock on the [`ITM`]. Should be called after any other mutating [`ITM`] functions.
```
tmplt, line 223:
```suggestion
    /// Engages the software lock on the [`ITM`]. Should be called after any other mutating [`ITM`] functions.
```

src/peripheral/itm.rs
```diff
@@ -174,42 +186,176 @@ pub struct ITMSettings {
...
+    /// [`LocalTimestampOptions::Disabled`] instead.
+    ///
+    /// *NOTE*: `GTSFREQ` and `SWOENA` in [`ITM_TCR`](struct@Tcr) fields
+    /// have potentially changed on this error.
+    TSPrescale,
+}
+
 impl ITM {
-    /// Removes the software lock on the ITM.
+    /// Removes the software lock on the [`ITM`]. Must be called before
```
tmplt, line 212:
```suggestion
    /// Disengages the software lock on the [`ITM`]. Must be called before
```

src/peripheral/itm.rs
```diff
@@ -174,42 +186,176 @@ pub struct ITMSettings {
...
+    /// Indicates whether the [`ITM`] is currently processing events.
+    /// Returns `true` if [`ITM`] events are present and are being drained.
     #[inline]
-    pub fn configure(&mut self, settings: ITMSettings) {
+    pub fn busy(&self) -> bool {
+        self.tcr.read().busy()
+    }
+
+    /// Configures the [`ITM`] with the passed [`ITMSettings`]. Returns `true`
+    /// if the configuration was successfully applied.
```
tmplt, line 258: This comment is outdated.

tmplt:
Only TraceBusID remains.

tmplt:
ARM DDI 0403E.d, p. C1-716:
> If multi-source trace is in use, the debugger must write a non-zero value to this field.

C-718:
> If a system supports multiple trace streams, the debugger must write a nonzero value to the ITM_TCR.TraceBusID field.
> For information about permitted values for this field in a CoreSight-compliant implementation, see the ARM IHI 0029E
> [...] An example of a system with multiple trace streams is an ARMv7-M core with ETM. In this case, the debugger must program the ETM TraceBusID register with a different nonzero identifier for the ETM trace stream.

ARM IHI 0029E does not contain the term "TraceBusID".

---

It is not very clear whether non-zero should be written if multi-source trace is *wanted*, or if the target *implements* it. If the latter, we'll need to write non-zero even if we only want a single trace source.

tmplt:
The refactored implementation of `ITM::configure` works as expected on the `atsame51n`. I've sent an email to ARM in hopes of some clarification on TraceBusID.

I consider this PR ready for review.

src/peripheral/dcb.rs
```diff
@@ -22,10 +24,7 @@ pub struct RegisterBlock {
 }
 
 impl DCB {
-    /// Enables TRACE. This is for example required by the
-    /// `peripheral::DWT` cycle counter to work properly.
-    /// As by STM documentation, this flag is not reset on
-    /// soft-reset, only on power reset.
+    /// Global enable for all [`DWT`] and [`ITM`] features.
```
TDHolmes, line 27: I think just doing a doc comment would be better than importing:
```suggestion
    /// Global enable for all [`DWT`](crate::peripheral::DWT) and [`ITM`](crate::peripheral::ITM) features.
```

src/peripheral/itm.rs
```diff
@@ -174,42 +186,177 @@ pub struct ITMSettings {
...
                 });
+
+                r
+            });
+        }
+        // GTSFREQ is potentially RAZ/WI
+        if settings.global_timestamps != GlobalTimestampOptions::Disabled
+            && self.tcr.read().gtsfreq() == 0
+        {
+            return Err(Error::GTS);
```
TDHolmes, line 299: should we re-lock in error case, or document that if there is an error ITM will be left unlocked?

src/peripheral/itm.rs
```diff
@@ -174,42 +186,177 @@ pub struct ITMSettings {
...
+
+    /// Checks whether the target implements the software lock
+    /// mechanism. If `true`, [`unlock`](ITM::unlock) must be called
+    /// before any other mutating [`ITM`] functions.
+    ///
+    /// See (coresight, B2.3.10).
+    #[inline]
+    pub fn has_software_lock(&self) -> bool {
+        self.lsr.read().sli()
+    }
```
TDHolmes, line 242: perhaps this should be checked for the user in `lock` and `unlock` so that if HW does support it, we `lock`/`unlock` and they can just call it without knowing. Similarly, `locked` could always return `false` if `has_software_lock` is false so it's all transparent.

src/peripheral/itm.rs
tmplt: I don't think there is any point in relocking on error. I'll document it.

src/peripheral/itm.rs
tmplt: Sounds good.

tmplt:
`lock` and `unlock` has been streamlined a bit and in the process `has_software_lock` and `locked` have been hidden. `lock` and `unlock` could potentially be hidden later when #393 and similar close so that `configure` is exhaustive.

src/peripheral/itm.rs
```diff
@@ -174,42 +186,188 @@ pub struct ITMSettings {
     pub timestamp_clk_src: TimestampClkSrc,
 }
 
+/// Possible errors on [ITM::configure].
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub enum ITMConfigurationError {
```
TDHolmes, line 191: perhaps `non_exhaustive`? Or are you sure these are the only errors `ITM::configure` will ever return?

TDHolmes (dismissed):
LGTM

src/peripheral/itm.rs
tmplt: Good catch. At least one more error enum will be added in the future.

tmplt:
Latest changes work as expected on hardware. Also tested on the `atsame51n`.

tmplt:
On TraceBusID: after a brief back-and-forth with ARM it *seems* to be valid to write non-zero. If only ITM is supported on the system, TraceBusID is potentially RAZ (in ARMv8-M, at least). Some rules do apply when multi-source trace is in use, but that's for another day.

tmplt:
This PR is ready for final review, @thalesfragoso.

tmplt:
The re-locking sequence currently incorrectly always assumes that a software lock is implemented. Amendmend will follow.

korken89:
Hi all,

What happened with this PR? Looking over it, it seems ready to go.

Ping @rust-embedded/cortex-m

thejpster (approved):
Looks ok to me, but I'm not an ITM expert

tmplt:
IIRC

> The re-locking sequence currently incorrectly always assumes that a software lock is implemented

is still relevant, and should be fixed before merge.

I do not have access to relevant hardware, and my knowledge of the area has faded somewhat. Is anyone else interested in taking up the torch?

korken89:
@tmplt I could probably give it a look. What HW is needed? I can get some ordered for testing.

tmplt:
I've tested the code on SAME51 and SAMV71 boards. IIRC one of these lack the software lock, resulting in a busy wait that never exits. I don't think the ITM coverage is exhaustively detailed in the data sheets, and so must be detected at run-time.

Issue #370, opened 2021-12-15
is ptr() function on `RegisterBlock`'s actually deprecated or not?


TDHolmes:
I see that the [`ptr() function`](https://github.com/rust-embedded/cortex-m/blob/ae1d2a62d895dc458784c465e021a010cb75a8b1/src/peripheral/mod.rs#L248) to get the pointer to the register block has the comment "to be deprecated in 0.7", but it's already 0.7 and it isn't deprecated... was this just missed or is it no longer the plan?

TDHolmes:
some context: https://github.com/rust-embedded/cortex-m/pull/235#issuecomment-662046920 / #235

adamgreig:
Oops, I think this was simply missed for 0.7. We can deprecate it in the existing 0.7 series though, and remove it for 0.8.

reitermarkus:
I think this can be closed, right?

PR #385, opened 2022-01-06
remove the ptr() function in favor of the PTR constant

TDHolmes:
Per #370/#235, the const fn ptr() was supposed to be deprecated and subsequently removed. This PR removes it for the master branch tracking the 0.8 release


adamgreig (approved):
Thanks for following this up!

bors merge

PR #387, opened 2022-01-08
Export semihistory::hio::open()

nviennot:
It's useful for opening various files on the host.

For example:
```rust
let mt file = open("file.bin\0", nr::open::RW_TRUNC_BINARY).unwrap();
file.write_all(b"Some binary content").unwrap();
```


jonathanpallant:
Sorry we dropped this one.

Is it still relevant? If so, consider taking a &CStr as the handle, which is the string-slice type for null terminated strings.

PR #415, opened 2022-01-23
Fix most clippy lints

newAM:
This fixes all clippy lints that do not result in a user-visible change, and result in valid code for the 1.40 MSRV.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -3,9 +3,9 @@ delete_merged_branches = true
 required_approvals = 1
 status = [
   "ci-linux (stable)",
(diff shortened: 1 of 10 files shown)


newAM:
Should be good now, I forgot `matches!` was introduced in 1.42.

adamgreig:
Since this is in the master branch for the 0.8 series, maybe it's worth bumping MSRV to 1.42 and getting `matches` in. I wouldn't be surprised if we end up wanting a newer MSRV for other 0.8 features anyway.

adamgreig (approved):
Thanks!

bors r+

PR #416, opened 2022-01-26
Fix #414, use fully qualified path for `exception` enum in `#[exception]` macro

### LINKED ISSUE #414: `cortex_m_rt::exception` is no longer hygenic in 0.7
Previously, one could apply `#[cortex_m_rt::exception]` fully qualified, as shown [here](https://github.com/cbiffle/lilos/blob/e4de81717b36a6fe4151f2db15a217829e9b998e/os/src/time.rs#L102).

With the release of 0.7.1, this has stopped working. For the macro to work, one must now `use cortex_m_rt::exception` and then use `#[exception]`. The error message is not great, but suggests that it's a problem with the macro referring to itself by an unqualified name internally:

```
error[E0433]: failed to resolve: use of undeclared crate or module `exception`
   --> src/time.rs:102:1
    |
102 | #[cortex_m_rt::exception]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `exception`
    |
    = note: this error originates in the attribute macro `cortex_m_rt::exception` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0433`.
```

(That message sent me on a bit of a goose chase figuring out why `exception` was not defined in `cortex_m_rt`, until I worked it out.)

adamgreig:
Partially reverts https://github.com/rust-embedded/cortex-m-rt/pull/224
to continue to use a fully-qualified path to `exception`.

### DIFF
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -328,7 +328,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             f.block.stmts = iter::once(
                 syn::parse2(quote! {{
                     // check that this exception actually exists
-                    exception::#ident;
+                    cortex_m_rt::exception::#ident;
                 }})
                 .unwrap(),
             )
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -174,7 +174,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
         Exception::Other => {
             quote! {
                 const _: () = {
-                    let _ = cortex_m_rt::Exception::#ident;
+                    let _ = ::cortex_m_rt::Exception::#ident;
                 };
             }
         }
@@ -328,7 +328,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             f.block.stmts = iter::once(
                 syn::parse2(quote! {{
                     // check that this exception actually exists
-                    cortex_m_rt::exception::#ident;
+                    ::cortex_m_rt::exception::#ident;
                 }})
                 .unwrap(),
             )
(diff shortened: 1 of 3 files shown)


cortex-m-rt/macros/src/lib.rs
```diff
@@ -328,7 +328,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
             f.block.stmts = iter::once(
                 syn::parse2(quote! {{
                     // check that this exception actually exists
-                    exception::#ident;
+                    cortex_m_rt::exception::#ident;
```
thalesfragoso, line 331: We should use `::cortex_m_rt` here. However, we are still on the 2015 edition here, right ? So we also have to add `extern crate cortex_m_rt`, are we inside a macro created module/function here ? Because if not, the `extern crate` might collide with another one inserted by the user.

adamgreig: Hmm, good point. We could bump to 2018? Our MSRV is already well beyond when 2018 came out.

thalesfragoso: Oh, the macros crate is actually 2018, is it okay to emit invalid 2015 code from a 2018 proc-macros crate ?

adamgreig: Incredibly it seems "yes": https://doc.rust-lang.org/edition-guide/editions/advanced-migrations.html#macro-hygiene

> Macros use a system called "edition hygiene" where the tokens within a macro are marked with which edition they come from. This allows external macros to be called from crates of varying editions without needing to worry about which edition it is called from.

Nevertheless I'd be in favour of bumping to 2018 edition for c-m-rt too...

thalesfragoso: Agreed.

thalesfragoso (approved):
Great, thanks again!

bors r+

PR #422, opened 2022-02-16
Add ITNS field to NVIC peripheral

sphw:
This PR adds the ITNS (Interrupt Target Non-secure) field to the NIVC peripheral. This field is required to write a boot loader for TrustZone-M devices, since it allows the user to use interrupts from non-secure states.

I believe I have maintained the correct padding for the next fields, but I have not tested these changes on a non M33 device. So a close review and test would be appreciated.

### DIFF
--- a/src/peripheral/nvic.rs
+++ b/src/peripheral/nvic.rs
@@ -33,17 +33,18 @@ pub struct RegisterBlock {
     /// Interrupt Active Bit (not present on Cortex-M0 variants)
     #[cfg(not(armv6m))]
     pub iabr: [RO<u32>; 16],
-    #[cfg(any(armv6m, armv8m))]
+    #[cfg(armv6m)]
     _reserved4: [u32; 16],
+
+    _reserved5: [u32; 16],
 
     #[cfg(armv8m)]
     /// Interrupt Target Non-secure (only present on Arm v8-M)
     pub itns: [RW<u32>; 16],
-    #[cfg(armv8m)]
-    _reserved5: [u32; 32],
-
     #[cfg(not(armv8m))]
-    _reserved5: [u32; 48],
+    _reserved6: [u32; 16],
+
+    _reserved7: [u32; 16],
 
     /// Interrupt Priority
     ///
@@ -74,7 +75,7 @@ pub struct RegisterBlock {
     pub ipr: [RW<u32>; 8],
 
     #[cfg(not(armv6m))]
-    _reserved6: [u32; 580],
+    _reserved8: [u32; 580],
 
     /// Software Trigger Interrupt
     #[cfg(not(armv6m))]
--- a/src/peripheral/nvic.rs
+++ b/src/peripheral/nvic.rs
@@ -33,9 +33,16 @@ pub struct RegisterBlock {
     /// Interrupt Active Bit (not present on Cortex-M0 variants)
     #[cfg(not(armv6m))]
     pub iabr: [RO<u32>; 16],
-    #[cfg(armv6m)]
+    #[cfg(any(armv6m, armv8m))]
     _reserved4: [u32; 16],
 
+    #[cfg(armv8m)]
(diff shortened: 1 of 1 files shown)


src/peripheral/nvic.rs
```diff
@@ -33,9 +33,16 @@ pub struct RegisterBlock {
     /// Interrupt Active Bit (not present on Cortex-M0 variants)
     #[cfg(not(armv6m))]
     pub iabr: [RO<u32>; 16],
-    #[cfg(armv6m)]
+    #[cfg(any(armv6m, armv8m))]
```
adamgreig, line 36: This change doesn't seem right to me - armv8m still has IABR right? And in any event the matching cfg above isn't changed, so armv8 would end up with an extra 16 byte offset here.

adamgreig, line 36: This change doesn't seem right to me - armv8m still has IABR right? And in any event the matching cfg above isn't changed, so armv8 would end up with an extra 16 byte offset here.

sphw: This doesn't disable IABR, it just re-uses the padding for both v6 `armv6m` and `armv8m`. I believe this is the correct padding. This is the ARM CMSIS definition for armv8m NVIC that has this padding

https://github.com/ARM-software/CMSIS_5/blob/a5743627a233e02388e583fbcfab29573b8fe643/CMSIS/Core/Include/core_armv8mml.h#L478-L481

adamgreig: I don't think it adds up right - on armv8 you'd have 16 words of IABR, then 16 words of padding, then 16 words of ITNS, then 32 words of padding, then IPR after that, which is 80 words in total instead of the correct 64.

For clarity I'd much rather have something like this, where every item is present or replaced by a same-sized padding:

```rist
    #[cfg(not(armv6m))]
    pub iabr: [RO<u32>; 16],
    #[cfg(armv6m)]
    _reserved4: [u32; 16],

    _reserved5: [u32; 16],

    #[cfg(armv8m))]
    pub itns: [RW<u32>; 16],
    #[cfg(not(armv8m)))]
    _reserved6: [u32; 16],

    _reserved7: [u32; 16],

    #[cfg(not(armv6m))]
    pub ipr: [RW<u8>; 496],
    #[cfg(armv6m)]
    pub ipr: [RW<u32>; 8],
```

Arguably we should write it as `cfg(any(armv7m, armv8m))` instead of `cfg(not(armv6m))`, though it might make it a bit trickier if we one day need to add something like armv9m.

sphw: Yeah you are right that it is off by 16 words of padding. I was confused by the fact that in the ARMv7 manual, IABR is specified as being 8 words long, but here it uses the ARMv7 register length of 16 words (presumably to support both platforms). I agree your method above would be cleaner. I'll change my PR to use that method.

sphw: Sorry for leaving this hanging, I just pushed a commit fixing the issue. LMK if there are any other problems

src/peripheral/nvic.rs
```diff
@@ -33,10 +33,18 @@ pub struct RegisterBlock {
     /// Interrupt Active Bit (not present on Cortex-M0 variants)
     #[cfg(not(armv6m))]
     pub iabr: [RO<u32>; 16],
-    #[cfg(armv6m)]
+    #[cfg(any(armv6m))]
```
reitermarkus, line 36: The `any` here is useless.

reitermarkus, line 36: The `any` here is useless.

sphw: fixed

macthestack:
Nice to see this! i was just about to submit a PR myself until I saw this.

Hopefully someone picks it up :)

adamgreig (approved):
Thanks for the reminder @macthestack! This looks good now, sorry for the wait after the first round of reviews @sphw!

bors r+

Issue #420, opened 2022-02-02
Tracking issue for stable inline assembly


adamgreig:
The next stable Rust release, 1.59, will include stable `asm!()` and `global_asm!()` macros that finally permit us to define assembly directly in the crate without needing to distribute precompiled binaries.

I anticipate this will let us remove all the current code to support precompiled binaries, including the outlined-inlined-asm in cortex-m and the `asm.S` in cortex-m-rt, and make all our `inline-asm` features no-ops. Since this shouldn't be a breaking change to users, it could be released as a cortex-m 0.7 backport, or as part of the upcoming 0.8, and as a normal minor 0.7 release for cortex-m-rt.

This issue is for tracking progress/concerns/problems.

hannobraun:
I just stumbled across this issue while doing some research. Can this be closed? I can't find any assembly in this repository that's not within `asm!` or `global_asm!`.

Issue #406, opened 2020-01-09
Make the `HardFault` trampoline opt-in


jonas-schievink:
Currently, *any* user of this crate has to pay for a small Assembly trampoline that provides access to the `&ExceptionFrame` to the `HardFault` handler (even if they don't use the `&ExceptionFrame`, or don't even define any `HardFault` handler):

https://github.com/rust-embedded/cortex-m-rt/blob/492c7bf6212486b378ecf0518572d180a50c83a1/asm.s#L8-L19

Since few people make use of the `&ExceptionFrame` argument, we could add a Cargo feature to turn it on instead, and generate the right linker script in `build.rs`. By forwarding the feature to the macro crate, `#[exception]` can provide good diagnostics when trying to use `&ExceptionFrame` without the feature enabled.

I am planning on extending the trampoline mechanism to other faults and maybe even all exceptions, which I've prototyped by making the ASM trampoline generic and have it work for all faults. This makes it larger and adds more data to the excutable, so making this opt-in would help reduce the cost. The trampoline code could also benefit from ARMv7/v8-specific improvements, which I'll work on later, which should shrink it again (for the higher platforms at least).

I suppose we could also have one Cargo feature per exception. The build script would then generate a linker script that pulls in exactly the trampolines that are needed. This should even still work with the precompiled `.a` files, if we do it right (putting every trampoline in its own section and making sure they're only pulled in when necessary).

Initially, defining 10 or 11 Cargo features for this sounds pretty sketchy, but most apps won't use any of them, and even "advanced" apps that implement an RTOS can get away with just 1 or 2 (one for SVCall to read syscall arguments, and one on HardFault/DefaultHandler for debugging crashes).

An ideal solution would not require Cargo features at all and instead make the `#[exception]` macro pull in the Assembly trampoline. I have not found any way to do this though. If one does come up (perhaps using the currently unstable [`#[link_args]`](https://github.com/rust-lang/rust/issues/29596) attribute), we can make the Cargo features no-ops until the next major version.

cbiffle:
@adamgreig - I believe this was closed inadvertently after being merely mentioned on #423. (Since I still can't work out how to turn off the trampoline without replacing the linker script.)

adamgreig:
I'm still seeing it as open, but it certainly shouldn't be closed yet, you're right.

[...]

Issue #265, opened 2020-09-04
SWI / SVC support


thejpster:
Could we add intrinsics for calling Software Interrupts (the SWI or SVC instruction)? We probably also want a mechanism for marshalling some number of integer arguments into registers, and back out again on the other side of the SWI (i.e. in the kernel).

TockOS do this with `llvm_asm!`, as you can see at https://github.com/tock/tock/blob/198ff4173e85f6b8474893c6784d38c12ead9dbb/arch/cortex-m0/src/lib.rs#L161

jonas-schievink:
As described in https://github.com/rust-embedded/cortex-m/issues/58, this is not in general possible since the specific `svc` ABI depends on the environment

thejpster:
Ah. I guess it would be unreasonable to have an array of 256 function pointers to let you pick the SVC value...

jonas-schievink:
Once inline assembly is stable, we can provide a const-generic function like this:

```rust
fn svc<const NUM: u8>() {
    unsafe {
        asm!("svc {imm}", imm = const NUM);
    }
}
```

thejpster:
> Once inline assembly is stable, we can provide a const-generic function like this:
> 
> ```rust
> fn svc<const NUM: u8>() {
>     unsafe {
>         asm!("svc {imm}", imm = const NUM);
>     }
> }
> ```

Nice!

thejpster:
So reading around a bit on this, some syscall ABIs just set the SVC value to be zero, as it's expensive to pull it out on the receiving side (you have to read the instruction that caused the interrupt, which will be in a different cache...). So, could we at least have a function which lets you SVC 0, and optionally populate some registers with arguments.

jamesmunns:
Just tagging this issue as I've hit this as well, and we now have stable asm.

That being said, I'm probably also just going to go the "svc 0" route as well.

adamgreig:
I guess we'll want https://doc.rust-lang.org/unstable-book/language-features/asm-const.html for this one.

korken89:
Seems to be working now :)
https://rust.godbolt.org/z/fPK5Mh5sz

Issue #215, opened 2020-05-05
add virtual unwind info (.debug_frame / CFI) to external assembly


japaric:
virtual unwinders (\*), for example based on `gimli::{DebugFrame,UnwindSection}`, are not able to unwind through external assembly subroutines like `__nop` because these subroutines lack [CFI](https://sourceware.org/binutils/docs/as/CFI-directives.html) (Call Frame Information), which -- when present -- gets encoded in the `.debug_frame` (\*\*) section of the ELF.

``` console
$ # no `.debug_frame` in the output
$ arm-none-eabi-size -Ax bin/thumbv6m-none-eabi.a
.text                 0x0    0x0
.data                 0x0    0x0
.bss                  0x0    0x0
(..)
.debug_line         0x160    0x0
.debug_info          0x22    0x0
.debug_abbrev        0x12    0x0
.debug_aranges       0xb0    0x0
.debug_str           0x2d    0x0
.debug_ranges        0xa8    0x0
.ARM.attributes      0x1c    0x0
(..)
```

For subroutines that do *not* touch the stack pointer or link register adding CFI with *no* register rules is sufficient to make virtual unwinding work. Adding that CFI looks like this:

``` diff
   .section .text.__nop
+  .cfi_sections .debug_frame
   .global __nop
   .thumb_func
+  .cfi_startproc
 __nop:
   bx lr
+  .cfi_endproc
   .size __nop, . - __nop
```

And yeah `bx lr` does not count as "touching LR" because it does not modify LR.

What needs to be done:

- [ ] the easy part: add this "no register rules" CFI prologue and epilogue to all assembly subroutines that do *not *modify the stack pointer (SP) or the linker register (LR) and then re-run `assemble.sh`. Some instructions that *do* modify SP / LR are `push`, `pop`, `stmia`, `ldmia`, `mov`/`msr` where SP/LR is the *destination* (first argument) -- do not modify subroutines that contain those instructions (see "hard part" below)

To test this part run the following command on the archives (`bin/*.a*`):

``` console
$ arm-none-eabi-readelf --debug-dump=frames thumbv7em-none-eabi.a
thumbv7em-none-eabi.a(asm.o)
Contents of the .debug_frame section:

00000000 0000000c ffffffff CIE
  Version:               1
  Augmentation:          ""
  Code alignment factor: 2
  Data alignment factor: -4
  Return address column: 14

  DW_CFA_def_cfa: r13 ofs 0

[...]

PR #423, opened 2022-02-22
Swap to just-stabilised asm!() and global_asm!() macros

adamgreig:
Once Rust 1.59 is released in a couple of days, the `asm!()` and `global_asm!()` macros will become stable, and we will no longer require the various precompiled binaries we've used until now in cortex-m and cortex-m-rt. cc #420.

This PR uses `asm!()` in cortex-m, and removes the inline-asm feature, since I anticipate this going into cortex-m 0.8 and so we don't need to leave it behind for compatibility. In various places the previous version would call an extern C method when built for the native target, which would only fail at link time; to preserve the ability to build on x86 I've either made the whole method require the `cortex_m` configuration, or where appropriate/convenient simply skipped the `asm!()` call.

This PR replaces the old gcc-preprocessed `asm.S` in cortex-m-rt with use of `global_asm!()`, although since you can't normally use `#[cfg(...)]` attributes with `global_asm!()`, there's also a slightly scary macro modified from one invented by @Dirbaio for a similar purpose. I considered putting the initialisation of LR behind an armv6m flag, but since we want to restore it after calling `__pre_init` it seemed better to just leave it the same on both targets. I added Cargo features to optionally set SP and VTOR at startup, which has been variously requested but would previously have required multiplicatively more pre-built binaries. Now: no problem. Relevant issues:

* https://github.com/rust-embedded/cortex-m-rt/issues/283
* https://github.com/rust-embedded/cortex-m-rt/issues/55
* https://github.com/rust-embedded/cortex-m-rt/issues/254
* https://github.com/rust-embedded/cortex-m-rt/issues/102
* https://github.com/rust-embedded/cortex-m-rt/pull/338

I've tested these on a couple of targets (and updated the CI): on the whole there's a small improvement in code size due to everyone getting inlined asm, especially in `cortex_m::interrupt::free()`.

The major downside is we bump our MSRV from 1.42 (March 2020) to 1.59 (Feb 2022). For cortex-m, I propose putting these changes in the upcoming 0.8 release (which is technically what the master branch is already on) and not backporting. For cortex-m-rt I'm not sure: we don't have any other pending breaking changes, so we could consider a patch release. Anyway, this PR doesn't commit to any particular releases, so we can decide that later.

For cortex-m-semihosting/panic-semihosting I think a patch release would be ideal, especially since we had to yank the last c-m-sh release due to conflicting prebuilt binaries (a problem that should now vanish).

Also tagging these issues that I think might also benefit from new inline asm:
* https://github.com/rust-embedded/cortex-m/issues/265
* https://github.com/rust-embedded/cortex-m/issues/215
* https://github.com/rust-embedded/cortex-m/issues/406


cortex-m-rt/src/lib.rs
```diff
@@ -431,8 +444,148 @@
 extern crate cortex_m_rt_macros as macros;
 
 use core::fmt;
+use core::arch::global_asm;
```
thejpster, line 447: I hope this clippy lint goes away in 1.59.

src/asm.rs
```diff
@@ -160,15 +209,17 @@ pub fn ttat(addr: *mut u32) -> u32 {
 #[inline]
 #[cfg(armv8m)]
 pub unsafe fn bx_ns(addr: u32) {
-    call_asm!(__bxns(addr: u32));
+    asm!("BXNS {}", in(reg) addr, options(nomem, nostack, preserves_flags));
```
thejpster, line 212: Why is this one all-caps?

thejpster:
Looks OK to me. Very excited to see this land!

adamgreig:
I've fixed that uppercase BXNS, set rust-version in the Cargo.toml (except the unchanged panic-itm), bumped those Cargos to edition 2021 while we're at it, and removed outdated references to old Rust versions from cortex-m-rt docs (since we now require 1.59 anyway, and they were already outdated with 1.42).

thejpster:
What happened to the CI?

Cargo.toml
```diff
@@ -29,7 +30,6 @@ optional = true
 [features]
 cm7 = []
 cm7-r0p1 = ["cm7"]
-inline-asm = []
```
Dirbaio, line 32: removing this is a breaking change, is the intent to release this as 0.8?

Perhaps it's better to leave it in with a note "deprecated, does nothing, kept for backwards compat"?

adamgreig: Yes, from top of thread: "For cortex-m, I propose putting these changes in the upcoming 0.8 release (which is technically what the master branch is already on) and not backporting."

adamgreig:
Whew, a few snags from the CI but all green now.

Cargo.toml
Dirbaio: The `inline-asm` features in the other crates could be removed then?

adamgreig: > For cortex-m-semihosting/panic-semihosting I think a patch release would be ideal, especially since we had to yank the last c-m-sh release due to conflicting prebuilt binaries (a problem that should now vanish).

But I'm sort of on the fence there, we could do remove the feature and do a breaking release. For cortex-m-semihosting we already released and yanked 0.4 and are bank on 0.3, so we could drop the feature and do 0.4 there. Perhaps that's nicest.

adamgreig: Pushed an update removing it from those crates too, may as well make a clean sweep of it.

thejpster (approved):

therealprof (approved):
LGTM

adamgreig:
bors r=thejpster

PR #424, opened 2022-03-01
Prepare for c-m-sh v0.4.2, p-sh v0.6.0

adamgreig:
This PR prepares for a release of cortex-m-semihosting 0.4.2 and panic-semihosting 0.6.0. Since we never had a non-yanked 0.4.x version of cortex-m-semihosting, this is essentially a new breaking release for both crates. The main change is the use of the new asm macro, with the removal of the inline-asm feature and bumped MSRV.

### DIFF
--- a/cortex-m-semihosting/CHANGELOG.md
+++ b/cortex-m-semihosting/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.5.0] - 2022-03-01
+
 - Always use inline-asm, requiring Rust 1.59.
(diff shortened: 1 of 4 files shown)


thejpster:
Although every 0.4.x was yanked, it did exist briefly and there is a risk someone somewhere might have it in their Cargo.lock. 0.5.0 is safer, and version numbers are free.

adamgreig:
Good point, updated to 0.5.

adamgreig:
Ping @rust-embedded/cortex-m, shall we get these released?

newAM (approved):
Looks good to me!

adamgreig:
bors r=newAM

PR #425, opened 2022-03-10
Remove prelude.

Dirbaio:
Hoping to sneak this in before 0.8 :)

The prelude just reexports all of `embedded-hal 0.2`'s prelude. With EH1.0 around the corner it would be a shame to get stuck with 0.2 in the prelude, and changing it later is a breaking change.

Also, EH1.0 has removed the prelude.


thejpster (approved):

newAM (approved):

therealprof (approved):
bors r+

PR #355, opened 2021-10-09
Add on-target tests to CI

newAM:
A week ago in the rust-embedded matrix chat @adamgreig mentioned that on-target CI would be helpful for the `cortex-m` and `cortex-m-rt` crates.  This is a pull-request to make that happen.

## History: Bootstrapping and `cortex-m-rt`

The first problem I came across was the amount of boilerplate required to build an on-target binary without `cortex-m-rt`.

There are two paths forward here:

1. Introduce a lot of boilerplate in `cortex-m`, which will largely be copy-pasted from `cortex-m-rt`.
2. Relocate `cortex-m-rt` to this workspace.

In (#391) `cortex-m-rt` was relocated to this workspace to fix the bootstrapping problem.

## Test Execution

Tests run with QEMU and physical hardware.

QEMU uses the LM3S6965 Cortex-M3 target because it is widely used.

The physical hardware is a NUCLEO-F070RB connected to a self-hosted runner.  In the future more hardware can be added to cover more CPUs.

Due to reliability concerns with self-hosted runners only QEMU will be a required status check in CI, the physical hardware checks will be informational only.

### CI Software

The CI software for running on physical hardware is simply a self-hosted github actions runner.  A working demonstration of this can be found [here](https://github.com/newAM/totally-not-a-cortex-m-fork/runs/5345451343?check_suite_focus=true) (the repository does not appear as a fork to work around github actions limitations with forks).

The runner uses [`probe-run`] to execute the tests on embedded hardware.  [`probe-run`] was chosen for several reasons:

* Actively maintained by [knurling-rs], with an actively maintained back-end from [`probe-rs`].
* Written in rust.  Understanding the code does not require contributors to learn a new language.
* Designed with on-target testing as a primary goal (for use with [`defmt-test`]).
* Automatic unwinding and backtrace display.

## Test Harness

This PR introduces a test harness, `minitest`.

`minitest` is almost identical to [`defmt-test`], the only difference is that it replaces [`defmt`] with [`rtt-target`] because [`defmt`] introduces a dependency cycle on `cortex-m`.

This is harness is very minimal, adding only 327 lines of rust:

```console
$ tokei testsuite/minitest 
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Markdown                1            7            0            4            3
 TOML                    2           41           34            0            7
-------------------------------------------------------------------------------
 Rust                    3          406          350            5           51
 |- Markdown             1            8            0            7            1
 (Total)                            414          350           12           52
[...]

PR #429, opened 2022-04-11
Exclude testsuite from cron, fixes #427

### LINKED ISSUE #427: CI Failure
Scheduled CI run failed. Details:

https://github.com/rust-embedded/cortex-m/actions/runs/2139988544

adamgreig:
Turns out #427 was caused by #355 after all, because the CI workflow was updated to exclude testsuite from its default `cargo test`, but the cron workflow wasn't.

For now this seems like the simplest fix; eventually it might be nice to have cron run the testsuite too (at least in qemu), but really cron's there to catch surprise compilation failures when new Rust versions or new dependency releases come out, so I don't think it's urgent.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -16,7 +16,7 @@ jobs:
           toolchain: stable
           override: true
       - name: Run tests
-        run: cargo test --all --exclude cortex-m-rt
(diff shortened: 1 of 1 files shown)


newAM (approved):
Look good to me!

therealprof (approved):
LGTM, thanks.

bors r+

PR #431, opened 2022-04-21
Add ability to set priority grouping.

reitermarkus:
Also update `interrupt::free` to pass `CriticalSection<'_>` instead of a reference.


src/peripheral/scb.rs
```diff
@@ -848,20 +867,39 @@ impl SCB {
...
         };
         crate::asm::dsb();
         loop {
             // wait for the reset
             crate::asm::nop(); // avoid rust-lang/rust#28728
         }
     }
+
+    /// Set the priority grouping.
+    pub fn set_priority_grouping(&mut self, grouping: PriorityGrouping) {
```
reitermarkus, line 884: Not sure if this is safe or whether it should be an `unsafe fn`.

reitermarkus:
@adamgreig, I extracted the unrelated changes from this PR. Should be easier to review now.

src/peripheral/scb.rs
```diff
@@ -119,6 +119,88 @@ mod fpu_consts {
...
+    /// Priority grouping 3
+    Prigroup3 = 3,
+    /// Priority grouping 4
+    Prigroup4 = 4,
+    /// Priority grouping 5
+    Prigroup5 = 5,
+    /// Priority grouping 6
+    Prigroup6 = 6,
+    /// Priority grouping 7
+    Prigroup7 = 7,
```
reitermarkus, line 143: I think these values are actually swapped, at least compared to the ST HAL.

reitermarkus, line 143: I think these values are actually swapped, at least compared to the ST HAL.

reitermarkus:
@adamgreig, ping.

PR #432, opened 2022-04-22
Fix inline assembly.

reitermarkus:
Fix some differences between previous `asm.S` and inline assembly.


adamgreig:
Thanks, good catch on the `cmp` especially. Is there any need for the `4:` label, do you think? I guess you already noticed we have to use `movs` instead of `mov` for that immediate load; the old asm.S loaded a literal from a constant pool but I think this way is a little quicker and takes a little less space.

reitermarkus:
> Is there any need for the `4:` label, do you think?

I guess not since it's not actually used.

adamgreig (approved):
Thanks!

bors merge

PR #436, opened 2022-05-04
Allow calling `interrupt::free` with zero-arity closures. (0.7.x)

reitermarkus:
0.8 will only accept zero-arity closures, this will make the transition a bit easier by being able to update calls before it is released.

### DIFF
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -56,6 +56,7 @@ pub unsafe fn enable() {
 /// https://geo-ant.github.io/blog/2021/rust-traits-and-variadic-functions/
 ///
 /// TODO: Remove before releasing 0.8.
+#[doc(hidden)]
 pub trait InterruptFreeFn<Args, R> {
     /// Call the closure.
     unsafe fn call(self) -> R;
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -48,20 +48,54 @@ pub unsafe fn enable() {
     call_asm!(__cpsie());
 }
 
+/// Hacky compatibility layer to allow calling `interrupt::free` using
+/// closures with arity 0 as well as 1. This trait is not considered
+/// part of the public API.
+///
+/// The generic `Args` type is not actually used, see:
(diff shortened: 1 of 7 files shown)


src/interrupt.rs
```diff
@@ -48,20 +48,54 @@ pub unsafe fn enable() {
...
 
+/// Hacky compatibility layer to allow calling `interrupt::free` using
+/// closures with arity 0 as well as 1. This trait is not considered
+/// part of the public API.
+///
+/// The generic `Args` type is not actually used, see:
+/// https://geo-ant.github.io/blog/2021/rust-traits-and-variadic-functions/
+///
+/// TODO: Remove before releasing 0.8.
+pub trait InterruptFreeFn<Args, R> {
```
newAM, line 59:
```suggestion
/// TODO: Remove before releasing 0.8.
#[doc(hidden)]
pub trait InterruptFreeFn<Args, R> {
```

Dirbaio:
This is unsound, it allows any value for the `'cs` lifetime. This code should be rejected:

```rust
    interrupt::free(|cs| {
        let cs: &'static CriticalSection = cs;
    });
```

Dirbaio:
The sound way of writiing it would be this:

```rust

impl<F, R> InterruptFreeFn<CriticalSection, R> for F
where
    F: for<'cs> FnOnce(&'cs CriticalSection) -> R,
{
    #[inline]
    unsafe fn call(self) -> R {
        self(&CriticalSection::new())
    }
}
```

However this fails to correctly infer the closure type:

```rust
interrupt::free(|cs| {})
```

Interestingly, these work.

```rust
interrupt::free(|cs: &CriticalSection| {})
interrupt::free(|cs: &_| {})
```

Closure type inference through "wrapper" traits like these is known to be flaky, see this issue for example https://github.com/rust-lang/rust/issues/70263 .  I'm not sure if what you're trying to do is possible in today's stable Rust.

reitermarkus:
> This is unsound

Thanks, good catch.

>  I'm not sure if what you're trying to do is possible in today's stable Rust.

Yeah, seems like it.

PR #437, opened 2022-05-05
Refactor `VectActive` to `Vector` and pass it to DefaultHandler.

reitermarkus:
Changes:

- Add a `Vector::map_interrupt` function to allow mapping the `Vector::Interrupt` variant to e.g. a HAL's interrupt type.
- Add a `SCB::vect_pending`function  for getting the pending vector, if any.


reitermarkus:
@@thalesfragoso, ping.

cortex-m-rt/examples/override-exception.rs
```diff
@@ -4,20 +4,18 @@
 #![no_main]
 #![no_std]
 
-extern crate cortex_m;
-extern crate cortex_m_rt as rt;
 extern crate panic_halt;
 
-use cortex_m::asm;
-use rt::{entry, exception, ExceptionFrame};
+use cortex_m::{asm, peripheral::scb::Vector};
```
jonathanpallant, line 9: I can't see `Vector` in https://docs.rs/cortex-m/latest/src/cortex_m/peripheral/scb.rs.html?search=Vector#19-98?

reitermarkus: What do you mean? I just added/renamed it in this PR.

jonathanpallant: Oh, brain fart. My bad.

jonathanpallant:
Sorry we dropped this one. It looks OK to me, but it's a breaking change.

@rust-embedded/arm how do we feel about breaking changes in cortex-m / cortex-m-rt 0.8?

src/peripheral/scb.rs
```diff
@@ -167,142 +168,196 @@ impl SCB {
...
+    ///
+    /// The return value is always within the closed range `[-14, -1]`.
+    #[inline]
+    pub const fn irqn(self) -> i8 {
+        self as i8
+    }
+}
+
+impl TryFrom<i8> for Exception {
+    type Error = i8;
```
thalesfragoso, line 281: Any reason for returning the argument back on error?

adamgreig:
> @rust-embedded/arm how do we feel about breaking changes in cortex-m / cortex-m-rt 0.8?

I don't have a problem with breaking changes in 0.8 per se, but breaking both c-m and c-m-rt at the same time in an inter-dependent way will make upgrading even harder than it already is, so the change would have to be really well motivated and I don't really understand the motivation here.

PR #439, opened 2022-05-06
fix bug on cmse::TestTarget::mpu_region

luojia65:
This function should read `mrvalid` and `mregion` fields from test target result, instead of `srvalid` and `sregion`.

### DIFF
--- a/src/cmse.rs
+++ b/src/cmse.rs
@@ -174,9 +174,9 @@ impl TestTarget {
     ///   * the TT instruction was executed from an unprivileged mode and the A flag was not specified.
     #[inline]
     pub fn mpu_region(self) -> Option<u8> {
-        if self.tt_resp.srvalid() {
-            // Cast is safe as SREGION field is defined on 8 bits.
-            Some(self.tt_resp.sregion() as u8)
+        if self.tt_resp.mrvalid() {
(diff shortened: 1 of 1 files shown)


thejpster:
Can you point to the spec that says this is how it should be?

luojia65:
@thejpster Thanks for noting! I found the official ARM documents here at: https://developer.arm.com/documentation/dui1095/a/The-Cortex-M23-Instruction-Set/Miscellaneous-instructions/TT--TTT--TTA--and-TTAT?lang=en

> Bits | Name | Description
> -- | -- | --
> [7:0] | MREGION | The MPU region that the address maps to. This field is 0 if MRVALID is 0.
> [15:8] | SREGION | The  SAU region that the address maps to. This field is only valid if the  instruction is executed from Secure state. This field is 0 if SRVALID is  0.
> [16] | MRVALID | Set to 1 if the MREGION content is valid. Set to 0 if the MREGION content is invalid.
> [17] | SRVALID | Set to 1 if the SREGION content is valid. Set to 0 if the SREGION content is invalid.

According to the documents, SREGION and SRVALID indicates the SAU region for this address; the MPU region is declared by MREGION and MRVALID.

thejpster:
Seems legit!

thejpster (approved):
bors r+

PR #441, opened 2022-05-15
Prepare for v0.7.5

adamgreig:
Currently with cortex-m 0.7.4 it's not possible for stable Rust users to enable the inline-asm feature, even though their compiler might support it, because of the `![cfg_attr(feature = "inline-asm", feature(asm))]` line. I propose a new 0.7.5 release that removes this line, which means users on stable Rust >=1.59 could enable the previously nightly-only feature to get stable inline asm.

I wouldn't enable the feature by default, because that would be a significant MSRV bump, but at least this way more users could enable it before we release 0.8 with the revamped and inline-only asm.

I've also backported the bugfix from #380 which I don't believe is a breaking change.

I haven't had a chance to test this yet so it would be great if someone could try it out and just make sure the inline-asm feature does work before merging.

Any thoughts on anything else worth backporting from 0.8?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -746,7 +746,8 @@ fn main() {
 - Functions to get the vector table
 - Wrappers over miscellaneous instructions like `bkpt`
 
-[Unreleased]: https://github.com/rust-embedded/cortex-m/compare/v0.7.4...HEAD
+[Unreleased]: https://github.com/rust-embedded/cortex-m/compare/v0.7.5...HEAD
+[v0.7.5]: https://github.com/rust-embedded/cortex-m/compare/v0.7.4...v0.7.5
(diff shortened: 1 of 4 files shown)


newAM (approved):
Looks good to me.  Didn't see anything else from 0.8 worth backporting.

adamgreig:
I've tested this and confirmed that code using one of the asm functions, when `inline-asm` feature is on and using release mode, gets the asm inlined directly into the call site, with a stable Rust compiler. Should be good to go!

thejpster:
So the only thing to be aware of is https://github.com/rust-embedded/cortex-m/issues/442 because if we get a PR for that, we'll almost certainly get a request to push it out in a new release.

adamgreig:
Since it looks like #442 relates to cortex-m-semihosting instead, I don't think it's a blocker for this.

PR #447, opened 2022-08-10
Add implementation for critical-section 1.0

Dirbaio:
Picking up #433 since it seems stalled. Changes from #433 are:
- Update to `critical-section 1.0.0-alpha.2`
- Use `bool` restore token
- Name Cargo feature `critical-section-single-core`.

TODO before merging:

- [x] Wait for `critical-section 1.0` release https://github.com/rust-embedded/critical-section/pull/19

Co-Authored-By: Markus Reiter @reitermarkus


.github/workflows/clippy.yml
```diff
@@ -1,19 +1,14 @@
 on:
   push:
     branches: [ staging, trying, master ]
-  pull_request_target:
+  pull_request:
```
adamgreig, line 4: What's the intention with changing this and removing the `pull_request_target` entries below? Won't it stop the clippy annotations working since the secret key becomes unavailable?

src/critical_section.rs
```diff
@@ -0,0 +1,29 @@
...
+    use crate::interrupt;
+    use crate::register::primask;
+
+    struct SingleCoreCriticalSection;
+    set_impl!(SingleCoreCriticalSection);
+
+    unsafe impl Impl for SingleCoreCriticalSection {
+        unsafe fn acquire() -> RawRestoreState {
+            let active = primask::read().is_active();
+            if active {
```
adamgreig, line 14: Do we need to condition disabling interrupts on them currently being active, or can we unconditionally disable them (while still returning whether they _were_ active for `release()`)?

.github/workflows/clippy.yml
Dirbaio: oops, probably a rebase mistake. Fixed.

src/critical_section.rs
Dirbaio: it's what the PR did, can't think of a reason either. Changed it to do it unconditionally.

reitermarkus:
> Picking up #433 since it seems stalled.

It's not stalled. I was just waiting for another review, same as my other open PRs. Feel free to ping me next time if you want me to rebase a PR.

Thanks anyways for picking this up.

The workflow `pull_request` change was meant to be temporary so CI can run on my fork.

src/interrupt.rs
```diff
@@ -54,20 +53,18 @@ pub unsafe fn enable() {
 }
 
 /// Execute closure `f` in an interrupt-free context.
```
adamgreig, line 55: Perhaps this method should now refer to critical-section, something like this:

```suggestion
/// Execute closure `f` in an interrupt-free context.
///
/// This method does not synchronise multiple cores and may disable required
/// interrupts on some platforms; see the critical-section crate for a cross-platform
/// way to enter a critical section which provides a `CriticalSection` token.
///
/// This crate provides an implementation for critical-section that simply disables
/// all interrupts which may be enabled with the `critical-section-single-core` feature.
```

Dirbaio: Done!

adamgreig (approved):
bors r+

adamgreig:
bors retry

should go after #450 merges...

PR #448, opened 2022-08-10
Add implementation for critical-section 1.0, for cortex-m v0.7.x

Dirbaio:
This is a subset of #447 without any breaking changes, just adding the `critical-section` implementation.

The goal is to release it in 0.7.6 so `critical-section` becomes usable without having to wait for cortex-m 0.8.

TODO before merging:

- [x] Wait for `critical-section 1.0` release https://github.com/rust-embedded/critical-section/pull/19

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,19 @@ jobs:
       matrix:
         # All generated code should be running on stable now
         rust: [stable]
+        features: ["", "critical-section-single-core"]
 
         include:
           # Test MSRV
           - rust: 1.38.0
+            features: ""
 
           # Test nightly but don't fail
           - rust: nightly
+            features: ""
+            experimental: true
+          - rust: nightly
+            features: "critical-section-single-core"
             experimental: true
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


Dirbaio:
Made `critical-section` an optional dep because it doesn't build for the 1.38 MSRV due to using `#[doc(include_str!(..))]`

reitermarkus:
So in 0.7.6, `interrupt::free` will keep using `bare_betal::CriticalSection` but the feature will allow other crates to already migrate to using `critical_section::with`, correct?

adamgreig:
> So in 0.7.6, `interrupt::free` will keep using `bare_betal::CriticalSection` but the feature will allow other crates to already migrate to using `critical_section::with`, correct?

That's right. We couldn't have `interrupt::free` use critical-section without breaking things unless we provided an impl by default, but doing so would stop people using a different impl on platforms that require them. Unless there's some clever trick to make `interrupt::free` on cortex-m 0.7.6 use critical-section with the single-core impl by default but still allow another impl to be defined...

Dirbaio:
I wouldn't make `interrupt::free` do something *other* than just disabling interrupts though. That'd be very confusing. Especially since in 0.8 it'll be back to just disabling interrupts (but not returning the CS token).

I'd just accept the fact that `interrupt::free` in 0.7 is unsound, and fix it in 0.8.

adamgreig (approved):
bors r+

PR #450, opened 2022-08-12
Fix CI compiletest breakage due to Rust 1.63

Dirbaio:


adamgreig (approved):
I thought this would break the MSRV tests, but it turns out we already only run the compiletest tests on stable, not MSRV or nightly.

bors r+

Issue #233, opened 2020-07-03
Critical sections do nothing in unprivileged code


cbiffle:
Currently, the `cortex_m::interrupt::free` mechanism, for executing a closure without the interference of interrupts, brackets the closure with `cpsid` and (conditionally, to support nesting) `cpsie` instructions.

These instructions do nothing in unprivileged mode. (You might expect them to trap; nope.) As a result, `cortex_m::interrupt:free` becomes a relatively expensive `FnOnce::call` in unprivileged code.

It's not clear whether `cortex_m` supports the privileged/unprivileged distinction (there aren't a lot of signs that it does, we're hitting a lot of issues) but this seems like a magnificent footgun.

Options for fixing this off the top of my head:

1. Use a Cargo feature to change the implementation of things that expect privileged code for use in unprivileged contexts. In this case, operations that are inherently privileged -- like all peripherals on the PPB -- would become contingent on the `privileged` feature and disappear in unprivileged code. Compile time failures are nice. This is my preferred approach, because my unprivileged code is separately compiled and memory isolated, but I bet it won't cover most peoples' use cases -- I suspect that most users of `cortex_m` that are doing any multitasking or priv/unpriv distinction are probably using one big blob of code without isolation, all linked against the same cargo features. (Also, because of how Cargo features are defined, it would be _super easy_ to accidentally pick up the `privileged` feature from your dependency graph. Laaame.)

2. Detect at runtime the current privileged mode and panic if `interrupt::free` is used in unprivileged code. This is a less-good option because it introduces a runtime failure, but it sure is easy.

3. Detect at runtime the current privileged mode and _call into hooks_ for entering and leaving a critical section. (There is no standard way of doing this from unpriv code; it depends on the OS. Our OS doesn't actually support global critical sections in unprivileged code for latency reasons, so we'd leave the hooks unimplemented if you went this way. I'm not even sure you could design a useful hook signature since `cortex_m::interrupt::free` is ambient authority that requires no context information.)

4. Declare that `cortex_m` is not to be used in code that uses unprivileged mode and hope that nobody ever does the wrong thing. (This is actually the conclusion we're heading toward; the APIs really do assume privilege. But it'd be nice to fix this.)

hug-dev:
[...]

PR #451, opened 2022-08-21
Small critical-section-related fixes.

Dirbaio:
See individual commit messages.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,16 @@
 //!
 //! # Optional features
 //!
+//! ## `critical-section-single-core`
+//!
+//! This feature enables a [`critical-section`](https://github.com/rust-embedded/critical-section)
+//! implementation suitable for single-core targets, based on disabling interrupts globally.
+//!
+//! It is **unsound** to enable it on multi-core targets or for code running in unprivileged mode,
+//! and may cause functional problems in systems where some interrupts must be not be disabled
+//! or critical sections are managed as part of an RTOS. In these cases, you should use
+//! a target-specific implementation instead, typically provided by a HAL or RTOS crate.
+//!
 //! ## `cm7-r0p1`
 //!
 //! This feature enables workarounds for errata found on Cortex-M7 chips with revision r0p1. Some
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,10 +49,6 @@ mod macros;
 pub mod asm;
 #[cfg(armv8m)]
 pub mod cmse;
-// This is only public so the `singleton` macro does not require depending on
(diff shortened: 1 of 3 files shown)


adamgreig:
While we're here, could you add the feature to the "Optional features" section of the docs in `src/lib.rs` please?

Dirbaio:
Done

thejpster (dismissed):
LGTM

src/lib.rs
```diff
@@ -9,6 +9,14 @@
 //!
 //! # Optional features
 //!
+//! ## `critical-section-single-core`
+//!
+//! This feature enables a [`critical-section`](https://github.com/rust-embedded/critical-section)
+//! implementation suitable for single-core targets, based on disabling interrupt globally.
```
adamgreig, line 15:
```suggestion
//! implementation suitable for single-core targets, based on disabling interrupts globally.
```

src/lib.rs
```diff
@@ -9,6 +9,14 @@
...
+//!
+//! It is **unsound** to enable it on multi-core targets. You should use a target-specific
+//! implementation that synchronizes all cores instead, typically provided by the HAL crate.
```
adamgreig, line 18:
```suggestion
//! It is **unsound** to enable it on multi-core targets or for code running in unprivileged mode,
//! and may cause functional problems in systems where some interrupts must be not be disabled
//! or critical sections are managed as part of an RTOS.
//! You should use a target-specific implementation instead, typically provided by a HAL crate.
```

src/macros.rs
```diff
@@ -31,7 +31,10 @@ macro_rules! iprintln {
 /// at most once in the whole lifetime of the program.
 ///
 /// # Notes
-/// This macro is unsound on multi core systems.
+///
+/// This macro requires a `critical-section` implementation to be set. For single core systems, you can
+/// enable the `critical-section-single-core` feature for this crate. For multi core systems, you
+/// have to provide one from elsewhere, typically your chip's HAL crate.
```
adamgreig, line 37:
```suggestion
/// This macro requires a `critical-section` implementation to be set. For most single core systems,
/// you can enable the `critical-section-single-core` feature for this crate. For other systems, you
/// have to provide one from elsewhere, typically your chip's HAL crate.
```

adamgreig (changes requested):
Few wording tweaks to highlight that even some single-core systems cannot use the feature, e.g. if unprivileged (see #233) or for things like softdevice or an RTOS.

adamgreig (approved):
thanks!

bors r+

adamgreig:
bors retry

PR #454, opened 2022-11-08
Prepare for cortex-m-rt 0.7.2

adamgreig:
It's been about a year since the last release so may as well get the couple of bug fixes and new assembly out, I think. This also adds the new set-vtor and set-sp features.


therealprof (approved):
SGTM, thanks.

bors r+

PR #462, opened 2023-01-26
panic-itm: update crate config so docs aren't empty

tgross35:
The docs are currently empty since the module is disabled https://docs.rs/panic-itm/0.4.2/panic_itm/index.html


newAM (approved):
Tested with `cargo doc` and `cargo rustdoc`, both work so this shouldn't break anything :)

Thanks for the fix!

bors r+

tgross35:
Thanks!

PR #464, opened 2023-02-11
cortex-m-rt: assert in linker script that stack_start is 8-byte aligned.

Dirbaio:
If the user sets RAM length to something that's not multiple of 8, the stack won't be 8-byte aligned. This'll trigger the same horrible symptoms as #463 .

This PR adds an assert to the linker script that enforces alignment.


adamgreig (approved):
Thanks!

bors r+

PR #467, opened 2023-02-13
cortex-m-rt: Remove LR push, to ensure the stack is 8-byte aligned.

Dirbaio:
This was causing incorrect execution of code optimized with the assumption the stack is 8-byte aligned.

Alternate version of #463
- Remove instead of fix the sentinel/fake frame.
- Remove code initializing LR, since it's now clobbered by the `bl main` anyway.
- ~~Remove the .cfi directives, since Reset now has no correct CFI info. I think this is the "correct" thing to do here.~~
- ~~Initialize the frame pointer in R7 (suggestion from @jamesmunns)~~

### DIFF
--- a/cortex-m-rt/src/lib.rs
+++ b/cortex-m-rt/src/lib.rs
@@ -507,12 +507,6 @@ cfg_global_asm! {
     ".cfi_startproc
      Reset:",
 
-    // Ensure LR is loaded with 0xFFFF_FFFF at startup to help debuggers find the first call frame.
-    // On ARMv6-M LR is not initialised at all, while other platforms should initialise it.
-    "movs r4, #0
-     mvns r4, r4
-     mov lr, r4",
-
     // If enabled, initialise the SP. This is normally initialised by the CPU itself or by a
     // bootloader, but some debuggers fail to set it when resetting the target, leading to
     // stack corruptions.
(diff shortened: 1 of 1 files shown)


cortex-m-rt/src/lib.rs
```diff
@@ -499,12 +499,10 @@ macro_rules! cfg_global_asm {
 // Finally jumps to the user main function.
 #[cfg(cortex_m)]
 cfg_global_asm! {
-    ".cfi_sections .debug_frame
```
jamesmunns, line 502: removing this made probe-run vomit for me, btw.

cortex-m-rt/src/lib.rs
```diff
@@ -572,20 +570,16 @@ cfg_global_asm! {
      dsb
      isb",
 
-    // Push `lr` to the stack for debuggers, to prevent them unwinding past Reset.
```
jamesmunns, line 575: This should be the only cfi/cfa directive you remove @Dirbaio

jamesmunns (approved):
Love it, :shipit:

adamgreig (approved):
On balance I want to go this route, instead of #463. Thanks for preparing all the PRs @Dirbaio, and thanks to everyone else who helped investigate, especially @Dirbaio and @jamesmunns, and @peter9477 for spotting the issue in the first place.

There's a lot of useful context, explanation, and research in the #463 thread, though, so future readers should check it out.

Apologies in advance to the probe-run team; I think in the end it's best to go directly to the solution we want to stick with, and hopefully it's not too bad to patch up probe-run to eliminate the warning it will start emitting.

bors r+

PR #468, opened 2023-02-14
cortex-m-rt: initialize the frame pointer chain.

Dirbaio:
AAPCS says the starting value should be 0: https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#6214the-frame-pointer

> The end of the frame record chain is indicated by the address zero in the address for the previous frame

### DIFF
--- a/cortex-m-rt/src/lib.rs
+++ b/cortex-m-rt/src/lib.rs
@@ -530,6 +530,11 @@ cfg_global_asm! {
      ldr r1, =__vector_table
      str r1, [r0]",
 
+    // Initialize the frame pointer chain.
+    // https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#6214the-frame-pointer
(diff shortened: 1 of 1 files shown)


adamgreig:
I'm not one to disagree with AAPCS, but I'm not sure there's any point including this extra instruction in the startup. It doesn't seem like any debugger even looks for this value, let alone does anything with a 0 value, and because no other startup code we've checked sets it, it seems very unlikely any debugger will start caring about it either. I agree in principle if every (or even most, or even any) startup reliably set FP=0 then a debugger could use this to help unwinding, but since debuggers don't do it, and startups don't do it, I'm not convinced we should be the ones to start doing it.

That said, I've not investigated debuggers besides gdb, and only the handful of startup codes in https://github.com/rust-embedded/cortex-m/pull/463#issuecomment-1426618145, so I'd welcome feedback that suggests this would actually be useful for someone. Otherwise it seems a waste, even if a very very small waste, to include it.

PR #483, opened 2023-07-11
Swap to GHMQ

adamgreig:
I'll update settings to remove bors as a required check and instead add the checks it used to require.


adamgreig:
The failure in nightly isn't related, so doesn't need to block this merging.

adamgreig:
Failure in nightly is reported upstream as https://github.com/rust-lang/rust/issues/113597

therealprof (approved):
LGTM

bors r+

PR #489, opened 2023-10-29
CI: actions-rs/toolchain -> dtolnay/rust-toolchain

newAM:
`actions-rs` is deprecated as of 2023-10-13, and was not maintained long before that.  This replaces all uses of [`actions-rs/toolchain`](https://github.com/actions-rs/toolchain) with [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

We still have a use of `actions-rs/clippy-check`.  For most repositories I simply run the CLI version, the UI isn't as nice, but it is easier to maintain.  I can make this change too, if desired.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,11 +25,9 @@ jobs:
             experimental: true
     steps:
       - uses: actions/checkout@v4
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          override: true
       - name: Run tests
         run: cargo test --all --exclude cortex-m-rt --exclude testsuite --features cortex-m/critical-section-single-core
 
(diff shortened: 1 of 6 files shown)


jonathanpallant:
Is the use of an action strictly required? I believe the standard Ubuntu environment has rustup installed, so you can just rustup whatever version of Rust you want.

jonathanpallant:
I think I'd still vote for having fewer dependencies in our CI actions and keeping things simpler (my word, that bash for parsing rust versions is utterly incomprehensible) but I'm not going to prevent anyone else coming along and approving this.

newAM:
I can see your point on using `rustup`.  However I'm using these actions for many repositories, and for that reason I would prefer to use it here as well since it is less work to maintain for me.

adamgreig (approved):
From scratch I'd go with rustup too, but I'm most interested in getting us moved off actions-rs and don't have any issue with the dtolnay action. If it does cause any issues it won't be hard to swap off it.

PR #496, opened 2023-11-19
Rename `__reset_vector` to `__exceptions`

### LINKED ISSUE #495: `__reset_vector` symbol placed incorrectly
`link.x.in` has this for placing the vector table members:

```
  .vector_table ORIGIN(FLASH) :
  {
    __vector_table = .;
 
    /* Initial Stack Pointer (SP) value. */
    LONG(_stack_start & 0xFFFFFFF8);
 
    /* Reset vector */
    KEEP(*(.vector_table.reset_vector));
    __reset_vector = .;
```

However when looking at the contents of the elf with `nm`, we see this at the start:

```
10000100 R __vector_table
10000104 00000004 R __RESET_VECTOR
10000108 00000038 R __EXCEPTIONS
10000108 R __reset_vector
10000140 00000080 R __INTERRUPTS
```

Here, `__RESET_VECTOR` is correctly placed at offset 4, but `__reset_vector` is placed at offset 8.

I tried adding an NMI (which should be at offset 8), and it was placed correctly, so I don't think there is any problem with layout, but the lowercase symbol is misleading.

Also this assert just looks wrong to me:

```
/* ## .vector_table */
ASSERT(__reset_vector == ADDR(.vector_table) + 0x8, "
BUG(cortex-m-rt): the reset vector is missing");
```

The vector table should be:

* 0x0: Stack Pointer
* 0x4: Reset Vector
* 0x8: NMI Vector
* ...

jamesmunns:
Closes #495

### DIFF
--- a/cortex-m-rt/link.x.in
+++ b/cortex-m-rt/link.x.in
@@ -79,11 +79,11 @@ SECTIONS
 
     /* Reset vector */
     KEEP(*(.vector_table.reset_vector)); /* this is the `__RESET_VECTOR` symbol */
-    __reset_vector = .;
 
     /* Exceptions */
+    __exceptions = .; /* start of exceptions */
     KEEP(*(.vector_table.exceptions)); /* this is the `__EXCEPTIONS` symbol */
-    __eexceptions = .;
+    __eexceptions = .; /* end of exceptions */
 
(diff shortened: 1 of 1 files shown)


adamgreig (approved):
thank you!

PR #500, opened 2023-12-14
Fixed VectActive::from() returning incorrect value

### LINKED ISSUE #499: Incorrect implementation of `cortex_m::peripheral::scb::VectActive::from()` in stable release 0.7.7
I discovered this bug accidentally while working with SCB and exceptions. `cortex_m::peripheral::scb::VectActive::from()` function returns incorrect value, which is 16 higher than the expected value. (in case when `VectActive` is `VectActive::Interrupt`.
I know this was already fixed in #332, but it was really long time ago (2 years ago) and this change is not present in the current released version uploaded on crates.io.

https://github.com/rust-embedded/cortex-m/blob/bb4a78208323260a161e68b2498438867f971bc5/src/peripheral/scb.rs#L303C15-L303C15

Since the fix is already implemented in the master branch, I am curious about the timeline for the next release on crates.io. Alternatively, is it feasible to release a patch for version 0.7.7 specifically to address this bug?

skibon02:
Fixes #499

Fixed `VectActive::from(u8)` now returns correct value for Interrupt variant
Refactored `SCB::vect_active()` to use `VectActive::from(u8)` and get rid of code duplication

### DIFF
--- a/src/peripheral/scb.rs
+++ b/src/peripheral/scb.rs
@@ -172,25 +172,8 @@ impl SCB {
     pub fn vect_active() -> VectActive {
         let icsr = unsafe { ptr::read(&(*SCB::PTR).icsr as *const _ as *const u32) };
 
-        match icsr as u8 {
-            0 => VectActive::ThreadMode,
-            2 => VectActive::Exception(Exception::NonMaskableInt),
-            3 => VectActive::Exception(Exception::HardFault),
-            #[cfg(not(armv6m))]
-            4 => VectActive::Exception(Exception::MemoryManagement),
-            #[cfg(not(armv6m))]
-            5 => VectActive::Exception(Exception::BusFault),
-            #[cfg(not(armv6m))]
-            6 => VectActive::Exception(Exception::UsageFault),
-            #[cfg(any(armv8m, native))]
-            7 => VectActive::Exception(Exception::SecureFault),
-            11 => VectActive::Exception(Exception::SVCall),
-            #[cfg(not(armv6m))]
-            12 => VectActive::Exception(Exception::DebugMonitor),
-            14 => VectActive::Exception(Exception::PendSV),
-            15 => VectActive::Exception(Exception::SysTick),
-            irqn => VectActive::Interrupt { irqn: irqn - 16 },
-        }
+        // NOTE(unsafe): Assume correctly selected target.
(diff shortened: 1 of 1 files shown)


newAM:
Can you rebase this?  The CI failure is from an old MSRV

skibon02:
@newAM, Could you please explain a little more? I forked the repo and made this commit from the v0.7.x branch.

newAM:
Oh, sorry, I didn't see that this was targeting `v0.7.x`!  That makes more sense then.  The problem is that `serde_json` isn't pinned, and the latest version requires edition 2021.  I think specify `>=1.00,<=1.0.100` for serde should fix the build failure.

skibon02:
Oh, `unwrap_unchecked` seems to not be available on rustc version `1.38.0`. It forces me to write a reaction to the case, when icsr register returns unexpected value. Is panic appropriate in this case? Or maybe we should execute `udf` instruction?

Assuming correct implementation for `VectActive::from` for all cortex-m processors, this problem can only appear when incorrect target was selected by user, and it is impossible to check this in compile time.

skibon02:
Adding these ugly lines in Cargo.toml to fix deps versions did help compiling on 1.38

Or should we better just bump MSRV to 1.60?

jonathanpallant:
cortex-m-rt says:

> This crate is guaranteed to compile on stable Rust 1.59.0 and up

skibon02:
But this is cortex-m, not cortex-m-rt
latest update 0.7.7 was published 1 year ago

From v0.7.x branch Readme:
https://github.com/rust-embedded/cortex-m/blob/bb4a78208323260a161e68b2498438867f971bc5/README.md?plain=1#L14

jonathanpallant:
But how many people use cortex-m without using cortex-m-rt?

skibon02:
cortex-m has twice as many downloads compared to cortex-m-rt
It seems that there isn't much sense in supporting these old versions of the Rust compiler. Moreover, a significant portion of people even uses Nightly. 

Should I bump the MSRV to 1.60 and remove fixed dependency versions?

jonathanpallant:
1.59 makes sense, so they are the same.

unwrap_unchecked first appeared in 1.58 so that's fine.

thejpster (approved):

PR #505, opened 2024-01-11
Fix nightly CI failures

newAM:
The `data_overflow` test seems to have been malfunctioning, and changing to `addr_of!` fixes it, but for some reason this test takes 10's of minutes to compile in CI?  I'm not sure what to make of this.


adamgreig:
Huh, the 10s of minutes sounds bizarre and kind of horrifying. I think the `addr_of` change is due to a new lint about taking the address of `static mut` variables (https://github.com/rust-lang/rust/pull/117556).

adamgreig (approved):

PR #510, opened 2024-01-30
cmrt: change condition on _stext location check

### LINKED ISSUE #509: More-helpful linker errors
Having just debugged an issue, it is clear that it would have helped me if the following asserts gave some clear information as to _why_/_how_ "The .text section must be placed inside the FLASH memory.". For example, add to this: "It is likely that this is failing because you are using more FLASH than you actually have"
https://github.com/rust-embedded/cortex-m/blob/6b3a5b7fb95fe98fb05ddb4270d2d14709afe99c/cortex-m-rt/link.x.in#L255-L262

adamgreig:
This avoids spurious errors when `.text` doesn't fit in `FLASH` which have nothing to do with the position of `_stext`. By not checking, we let the linker provide its own, more-useful, error.

Closes #509.

### DIFF
--- a/cortex-m-rt/link.x.in
+++ b/cortex-m-rt/link.x.in
@@ -257,9 +257,9 @@ ASSERT(ADDR(.vector_table) + SIZEOF(.vector_table) <= _stext, "
 ERROR(cortex-m-rt): The .text section can't be placed inside the .vector_table section
 Set _stext to an address greater than the end of .vector_table (See output of `nm`)");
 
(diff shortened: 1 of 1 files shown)


thejpster (commented):
This removes the + sizeof(.text)?

thejpster:
Ah, that's by design. https://github.com/rust-embedded/cortex-m/issues/509#issuecomment-1917771959

thejpster (approved):

PR #515, opened 2024-03-13
HardFaultTrampoline now passes &mut ExceptionFrame

SCingolani:
In order to allow for a user provided HardFault handler to modify the exception frame (e.g. to return to a different PC) the signature for the function to be used with the `#[HardFault]` handler now has a `&mut ExceptionFrame` instead of just a reference.

Use case: we want to catch an exception, so we set some flag, run some code that might trigger a HardFault, update the flag in the handler, and return to the instruction _after_ the one that caused the exception. We can then check the flag to see if an exception occurred or not. Key point is: to return to the instruction following the one that caused the exception, we want to modify the PC on the `ExceptionFrame` before loading EXC_RETURN onto the PC.


thejpster:
This would be a big breaking change.

Can the macro be adjusted to accept either a shared reference or a mutable reference as an argument to the handler function?

PR #518, opened 2024-03-23
Use correct cfg for `cm7` feature

### LINKED ISSUE #517: Incorrect usage of `cm7` feature
I believe in #352, the incorrect `#[cfg(...)]` was used for the `cm7` feature.   The `cm7` feature was added but it doesn't actually activate anything because `#[cfg(cm7)]` instead of `#[cfg(feature = "cm7")]`
For example:
https://github.com/rust-embedded/cortex-m/blob/549e951d55e0d90141ac92069f04d3b5c775edd3/cortex-m/src/peripheral/mod.rs#L95-L98

This is fairly easy to workaround at the moment by simply defining the cfg either by:
1. `println!("cargo:rustc-cfg=cm7")`
    -  in the build.rs script
2. `rustflags = [ "--cfg", "cm7"]`
    - in Cargo  config.toml 

Maybe this behavior was changed in some Rust version and I'm not aware of it,  but at least for me with the "rustc 1.78.0-nightly (f4b771bf1 2024-03-14)"  version its how I described and by activating the `cm7` feature it fails to compile when trying to access the `AC` field or module.

Fix is very simply,   just replacing `#[cfg(cm7)]`   ->  `#[cfg(feature = "cm7")]`,  I can open PR for it, would just need to figure out what's the PR process here :)

Almeida-Oco:
Fixes #517.

Change all occurrences of `#[cfg(cm7)]` into `#[cfg(feature = "cm7")]` so the code sections are actually accessible with the `cm7`  feature.

### DIFF
--- a/cortex-m/src/peripheral/mod.rs
+++ b/cortex-m/src/peripheral/mod.rs
@@ -60,7 +60,7 @@
 use core::marker::PhantomData;
 use core::ops;
 
-#[cfg(cm7)]
+#[cfg(feature = "cm7")]
 pub mod ac;
 #[cfg(not(armv6m))]
 pub mod cbp;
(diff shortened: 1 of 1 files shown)


adamgreig (approved):
thanks for catching and fixing this!

PR #522, opened 2024-04-02
singleton: forward attributes

### LINKED ISSUE #521: forward attributes through `singleton!()`
Especially for DMA buffers correct placement in dedicated linker section and memories is crucial (e.g. ethernet dma descriptor rings). It would be great if `singleton!()` would support attaching attributes.

jordens:
This only works in the long form with a named variable because of
constraints on where the `vis` metavariable can be used.

close #521


jordens:
The visibility thing might be useless as this ends up in a closure. I'm open to input.

Dirbaio (approved):
lgtm! nice

adamgreig:
Supporting attributes looks good and is definitely really useful. I can't tell when the visibility spec could ever make a difference though - the only way anything outside of the macro itself can refer to the storage is by the `&mut` returned from the macro, so I don't think it's material whether the underlying static is `pub` or not. Unless there's some other compiler check I'm missing about how a reference to a static gets used?

adamgreig:
Oh, true, that wasn't possible before but I suppose it is now. I think ideally the static should _not_ be accessible from outside the macro... but using an `export_name` attribute and separate `extern` already requires `unsafe` so perhaps we can allow that.

I don't think there's any point including the visibility specifier as it seems like it can never make any difference, so I'd rather remove it.

jordens:
@adamgreig removed.

therealprof (approved):
LGTM

adamgreig (approved):

PR #525, opened 2024-04-16
Prepare for cortex-m-rt v0.7.4

adamgreig:
Draft while we think about updating some documentation too.


cortex-m-rt/CHANGELOG.md
```diff
@@ -7,10 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 
-- Add `zero-init-ram` feature to initialize RAM with zeros on startup. This can be necessary on
-  safety-critical hardware to properly initialize memory integrity measures.
-- Add optional `exception` argument for `HardFault`. It has one option `trampoline` which is true by default. When set to false, no trampoline will be created and the function will be called as the exception handler directly.
+## [v0.7.4]
+
+- Add `zero-init-ram` feature to initialize RAM with zeros on startup.
+  This can be necessary to properly initialise hardware error detection.
+- Add optional `exception` argument for `HardFault`.
+  It has one option `trampoline` which is true by default. When set to false,
```
therealprof, line 15:
```suggestion
  It has one option `trampoline` which is `true` by default. When set to `false`,
```

cortex-m-rt/CHANGELOG.md
```diff
@@ -7,10 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  no trampoline will be created and the function will be called as the
+  exception handler directly.
 - MSRV increased to 1.60.0 to align with `embedded-hal` version 1.
```
Dirbaio, line 18: cmrt doesn't use embedded-hal. do we track the msrv per-crate, or are we ok with bumping it because we unify it for the whole repo?

adamgreig: it's just been bumped to unify it for the whole repo, but there's no pressing need to actually do that since it's a separate CI run, I'll see if we can revert it.

adamgreig: ah, our tests/examples depend on cortex-m, so the CI ends up needing 1.60 anyway...

adamgreig: I think it's simplest to leave it at 1.60 and unified with cortex-m.

therealprof (approved):
LGTM

PR #526, opened 2024-04-16
Remove nomem flag from isb, dsb and dmb

### LINKED ISSUE #502: Is `nomem` a correct option for `isb`, `dsb` and `dmb`?
In the nightly reference for inline assembly it says for `nomem` that: "The compiler also assumes that this asm! block does not perform any kind of synchronization with other threads, e.g. via fences."
The same wording is not in the stable docs but I don't know if it is only added due to doc update or if there is an actual difference in how the compiler works.
Should those fences have the `nomem` option set?

jannic:
Fixes #502


adamgreig (approved):
thanks!

PR #533, opened 2024-05-19
Prepare for cortex-m-rt 0.7.5

### LINKED ISSUE #532: HardFault handler is completely broken in c-m-rt 0.7.4
cortex-m-rt 0.7.4 changes the exception macro so that hardfaults now have an *optional* trampoline argument. If the argument isn't given, the exception frame reference isn't passed.

Unfortunately the macro crate wasn't published, and cortex-m-rt has a tight pin on the old version: 0.7.0

This means if you use the exception frame argument, the reference will get a garbage value. This causes a failure *inside the hard fault handler*, putting the CPU into a lock-up state.

We need to publish the cortex-m-rt-macros crate, yank cortex-m-rt 0.7.4, and publish 0.7.5 with a pin on the new macro crate. And then somehow update the docs / write some tests so that in future it's much more difficult to release cortex-m-rt with the wrong version of the macro crate.

adamgreig:
Fixes #532.

### DIFF
--- a/cortex-m-rt/CHANGELOG.md
+++ b/cortex-m-rt/CHANGELOG.md
@@ -7,8 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.7.5]
+
+- Fix incorrect dependency on cortex-m-rt-macros in v0.7.4 which led to
+  incorrect HardFault handlers being generated.
 - MSRV is now Rust 1.61 to support syn verions >=2.0.68
-- The `_stack_end` symbol is added, e.g. for use with MSPLIM.
+- The `_stack_end` symbol is added, e.g. for use with MSPLIM (#565)
+- Macros now expand to include an `unsafe` block and `allow(static_mut_refs)`
+  when generating the `&'static mut` references in entry/interrupt handlers
+  (#561, #558)
+- `expect` attribute now permitted on entry/interrupt handlers (#557)
+- Up to 496 interrupts now permitted on ARMv8-M (#555)
+- Add new `paint-stack` feature which writes a fixed value to the entire stack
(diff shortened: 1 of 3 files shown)


thejpster (approved):

thejpster:
Before we do the releases, have we checked that the new macro works as expected and rejects mismatched hard fault signatures?

diondokter:
> Before we do the releases, have we checked that the new macro works as expected and rejects mismatched hard fault signatures?

Tests were added: https://github.com/rust-embedded/cortex-m/blob/master/cortex-m-rt/tests/compile-fail/hard-fault-bad-signature-3.rs
There are more too.

It doesn't pick up on unreleased macro crate releases because it defines the dependency with a path:
https://github.com/rust-embedded/cortex-m/blob/4395baee03d57a4e1763c427073da5f33ec3b17e/cortex-m-rt/Cargo.toml#L21-L22
This is probably the correct way in this crate, but yeah it does make it easy to forget to release any new version of it.

jonathanpallant:
I think one solution for this is for the main branch to be changed to have the version number `N + 1` right after version `N` is released, and for any dependencies within the same repo to also use `N + 1` as a version number.

jamesmunns:
Also noting that @newAM successfully yanked v0.7.4 - we should probably note this in the release docs that we are fixing those regression(s).

newAM:
I rebased this branch to fix CI and resolve conflicts.

newAM (approved):
I missed that this PR got blocked await a review after thejpster left as a member.  I'm not sure how to evaluate the risk of #535, will wait for other cortex-m members to help evaluate that.

adamgreig:
I've rebased and updated the changelog with the various changes. Should be ready to go now.

newAM (approved):

PR #534, opened 2024-05-21
CI: select probe for on-target test

newAM:


adamgreig (approved):

adamgreig:
Ugh, it looks like a `syn` minor version bump increased their MSRV to 1.61, beyond our current 1.60...

newAM:
I attempted to pin `syn`, with something similar to:

```toml
[dependencies.syn]
version = ">=2.0.0,<2.0.53"
optional = true

[features]
serde = ["dep:serde", "dep:syn"]
```

However our MSRV doesn't support `dep:` either.

I could pin our lockfile in CI, but that doesn't solve the problem for downstream consumers.

I think the easiest way out for now is an MSRV bump?

adamgreig:
Yep, I think go ahead with an MSRV bump in this PR.

adamgreig (approved):

PR #539, opened 2024-06-23
Expose configuration of the SEVONPEND bit within the System Control Register (SCR)

calebstewart:
This PR exposes the `SEVONPEND` bit of the System Control Register (`SCR`) within the SCB type. This seems inline with the current implementation of the other two defined `SCR` fields: `SLEEPDEEP` and `SLEEPONEXIT`. 

As a side-note, I'm relatively new to embedded development in rust, so if this has been omitted for some specific reason I'm unaware of, I'm happy to be corrected. :) That being said, I simply implemented these two new methods in the same way that `set/clear_sleepdeep` and `set/clear_sleeponexit` were implemented with the help of section `B3.2.7` of the [ARMv7-M Architecture Reference Manual](https://developer.arm.com/documentation/ddi0403/latest/).

### DIFF
--- a/cortex-m/src/peripheral/scb.rs
+++ b/cortex-m/src/peripheral/scb.rs
@@ -832,6 +832,26 @@ impl SCB {
     }
 }
 
+const SCB_SCR_SEVONPEND: u32 = 0x1 << 4;
+
+impl SCB {
+    /// Set the SEVONPEND bit in the SCR register
+    #[inline]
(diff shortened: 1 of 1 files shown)


newAM (approved):
Thanks!

newAM:
Re-queued this to allow #542 to go through first which speeds up CI (and the merge queues) considerably.

PR #541, opened 2024-06-30
CI: reduce duration

newAM:
The data_overflow test takes 3 minutes to compile on my system. On the GitHub actions runners it takes 10-15 minutes. I do not know the underlying reason for this.

Splitting up the arrays in data_overflow into smaller 8k chunks does not change the test outcome, but rustc compiles it in 5 seconds (36x faster).


newAM:
There are some problems with this... It does change the outcome, but ci/script.sh doesn't check the status code.

Going back to the drawing board on why this takes so long.

Issue #404, opened 2019-11-21
`data_overflow.rs` example/test is failing


jonas-schievink:
Or rather, it builds successfully. The reason is that these `!` do not work with `errexit`, so a successful build has no consequences:

https://github.com/rust-embedded/cortex-m-rt/blob/5ac023dd9261d67aafbeab018241b4bcc4742152/ci/script.sh#L44-L48

Caught by shellcheck [SC2251](https://github.com/koalaman/shellcheck/wiki/SC2251).

newAM:
I'm triaging Cortex-M issues, I think this one is fixed as of #505, please re-open if not!

newAM:
Actually I didn't fix it in #505, have #542 open to fix.

PR #542, opened 2024-06-30
CI: fix data_overflow test

newAM:
Long story:

* This test was introduced
* `ci/script.sh` did not correctly check for failures, [SC2251]
  * Thanks #404 for the tip!
* The FLASH size was increased, and the test incorrectly passed, but nobody noticed.
* I modified the test in #505 which made it fail again, but for the wrong reason.  `ptr::read_volatile(ptr::addr_of!(RODATA))` reads the entire array, which is not equivalent to the original code `ptr::read_volatile(&RODATA as *const u8)` which read a single element of the array.
* The test now failed, but the stack related overflow takes rustc a LONG time to compile, and pushed our CI times to >30m.

These changes fix ci/scripts.sh to exit with a non-zero code if data_overflow is passing, and makes data_overflow fail to compile for the original reason, updating RODATA to reflect the increased FLASH size.

[SC2251]: https://www.shellcheck.net/wiki/SC2251

### DIFF
--- a/cortex-m-rt/ci/script.sh
+++ b/cortex-m-rt/ci/script.sh
@@ -55,8 +55,8 @@ main() {
                 cargo rustc --target "$TARGET" --example "$ex" --features "${needed_features}" --release -- $linker
             done
             for ex in "${fail_examples[@]}"; do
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
thanks for figuring this one out!

PR #548, opened 2024-08-07
Adds a `paint-stack` feature.

thejpster:
Paints the stack with a fixed 0xcccccccc value, which makes it easier to see the stack high-water mark.

Also update the qemu.rs example to verify a .data value and a .bss value.

### DIFF
--- a/cortex-m-rt/Cargo.toml
+++ b/cortex-m-rt/Cargo.toml
@@ -46,6 +46,7 @@ device = []
 set-sp = []
 set-vtor = []
 zero-init-ram = []
+paint-stack = []
 
 [package.metadata.docs.rs]
 features = ["device"]
--- a/cortex-m-rt/examples/qemu.rs
+++ b/cortex-m-rt/examples/qemu.rs
@@ -1,7 +1,14 @@
 #![no_main]
 #![no_std]
 
-use core::fmt::Write;
+use core::{
+    fmt::Write,
+    sync::atomic::{AtomicU32, Ordering},
+};
+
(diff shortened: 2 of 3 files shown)


adamgreig:
Nice idea and implementation, thanks! Could you fix the clippy errors?

I'm also not entirely sure it's worth having `STACK_PAINT_VALUE` especially if we can't even refer to it from the assembly. In other words, maybe we just document that the stack is painted with 0xCCCCCCCC and put that literal in the asm. On the other hand, if we can one day refer to the const in the asm block maybe that will be nicer.

I think either way we should write the actual paint value in the documentation too, because otherwise the first thing anyone has to do is resolve the value of that constant. It's not like we'll ever have to change it either (...right?).

thejpster:
The constant was for whoever writes a tide mark function in the future, and so it appears in the docs.

The clippy failures are due to new lints in stable Rust, not my changes. You might want another PR that fixes them all, to keep this one clean. Or I can try and fix them.

adamgreig:
I've fixed the CI but you'll have to rebase the PR for it to pass, would you mind? You can also revert the spacing changes needed to appease clippy as they're already in master (but hopefully that should happen automatically when you rebase).

> The constant was for whoever writes a tide mark function in the future, and so it appears in the docs.

OK, makes sense. Could you mention the value of the constant in the docs rather than only having it by name?

adamgreig:
Thanks! I pushed a small change to mention the value in the docs and change the note about keeping it in sync from the docstring to a regular comment, if you think that looks OK I'll go ahead and merge.

adamgreig (approved):

PR #555, opened 2024-08-28
Allow up to 496 interrupts on ARMv8-M

### LINKED ISSUE #529: Not enough interrupts
> In the Armv8-M architecture, the NVIC supports up to 496 general-purpose interrupt lines. However, processor implementations may limit the maximum to a lower number, often 240 or 480. Cortex-M processors let system designers choose any number of general-purpose interrupt lines appropriate to their system, up to the specified limit. In Cortex-M processors that include the Mainline extension, system designers can choose the number of bits implemented in the programmable priority value for each exception.

-- <https://documentation-service.arm.com/static/64c7832738511951cb7a246e>

We only allow up to 240 interrupts in this crate, but Armv8-M allows up to 496 (plus 16 exceptions to make 512 table vector entries in total).

Tiwalun:
Just run into this while trying to generate a PAC, this fixes #529 .

Not sure if it would be better to change the default value to 496 as well.

### DIFF
--- a/cortex-m-rt/build.rs
+++ b/cortex-m-rt/build.rs
@@ -59,7 +59,7 @@ INCLUDE device.x"#
     } else if target.starts_with("thumbv8m") {
         println!("cargo:rustc-cfg=cortex_m");
         println!("cargo:rustc-cfg=armv8m");
-        240
+        496
     } else {
         // Non ARM target. We assume you're just testing the syntax.
         // This value seems as good as any.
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
Thanks, looks good! Just noting that this changes the size of the generated interrupt table (thus taking up appreciably more flash) only when the `device` feature is not used, which is rare - for most people the table comes from the PAC and so this won't impact flash size. Instead it just stops cortex-m-rt rejecting perfectly valid device crates with lots of interrupts.

PR #556, opened 2024-09-01
CI: remove self-hosted runner

newAM:
I'm moving soon, and will be traveling for a bit while I'm between places.  Sadly I don't have a place to setup the STM32 board while I'm away.

For the basic smoke tests that exist QEMU is more than sufficient coverage, I sadly still haven't gotten around to adding more meaningful tests.


adamgreig (approved):

adamgreig:
thanks!

PR #558, opened 2024-09-16
Wrap getting a &mut from a static mut in an unsafe block

### LINKED ISSUE #537: `static mut` transform produces warning with unsafe-op-in-unsafe-fn
```rust
#![warn(unsafe_op_in_unsafe_fn)]
#[entry]
fn main() -> ! {
    // The next line generates a warning while it shouldn't.
    static mut CLOCKS: MaybeUninit<Clocks> = MaybeUninit::new();
    // ...
}
```

I see in https://github.com/rust-embedded/cortex-m/issues/411#issuecomment-784449125 that maybe this `static mut` transform is going to be deprecated. Is that true?

jannic:
Fixes #537

### DIFF
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -85,7 +85,7 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
                 {
                     #(#attrs)*
                     static mut #ident: #ty = #expr;
-                    &mut #ident
+                    unsafe { &mut #ident }
                 }
             }
         })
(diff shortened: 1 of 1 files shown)


ia0:
Note that we'll probably still get the `static-mut-refs` lint right? (i.e. this only fixes the `unsafe-op-in-unsafe-fn` lint)

Not sure what's the best: just allow the lint or try to use `&raw mut` (but that might trigger other lints `deref-addrof`).

ia0:
> That's already taken care of by #561

Perfect! I didn't see it since it's not released yet and got the warning with a recent compiler.

adamgreig (approved):

Issue #411, opened 2021-02-23
statics defined in #[entry] are not safe on multi-core systems


Disasm:
RP2040 starts execution on both cores. If entry function contains a "special" definition of `static mut` variable, each core takes a mutable reference to it, which is illegal from the Rust point of view.

Here is the relevant piece of `#[entry]` code:
https://github.com/rust-embedded/cortex-m-rt/blob/ca4790f40738bcc770285c8080955a2077682078/macros/src/lib.rs#L87-L88

thejpster:
It's unsound where the interrupt function can be re-entered. The NVIC usually prohibits this but on a multi-core system you need something at the HAL level to ensure this. You can imagine a mechanism that moves ownership of system interrupts from one core to another, hiding the regular NVIC API from the user. Then it would still be sound.

9names:
> The NVIC usually prohibits this but on a multi-core system you need something at the HAL level to ensure this.

You can also re-enable interrupts inside an interrupt and get the same problem.

In spite of rp2040 being around for several years we haven't had a single report of anyone having issues with this.

Rather than trying to make this safe, I think we should just document that you should not do anything that allows concurrent/recurrent execution of interrupt handlers and leave it at that, since the actual real-world usages of this behaviour are exceptionally rare, and even in the absence of the static mut transform are likely to be unsafe.

jannic:
> In spite of rp2040 being around for several years we haven't had a single report of anyone having issues with this.

I'm not sure if this is a strong argument. Having two instances of `&mut` references to the same underlying static is instant UB, but that doesn't guarantee that something bad happens. And even if it somehow breaks the software, it may happen very occasional, basically impossible to reproduce, and with non-obvious symptoms. So it may be dismissed as a flaky device and never be traced back to the underlying cause.

IMHO if we know that there are conditions that the user must hold up to avoid UB, there should be at least one `unsafe` function with a useful safety comment, so the user has a fair chance of knowing about these requirements, and a point to start looking for UB in case the software doesn't work as expected.

thejpster:
It should go behind a flag, like the cs impl. Maybe even the same flag.

PR #561, opened 2024-10-12
Allow static_mut_refs lint in macro expansions

jannic:
On ~~nightly~~ beta, the static_mut_refs lint defaults to warn. With some of our examples using `#[deny(warnings)]`, this leads to the following error:

```
error: creating a mutable reference to mutable static is discouraged
  --> cortex-m-rt/examples/entry-static.rs:15:16
   |
15 |     static mut COUNT: u32 = 0;
   |                ^^^^^ mutable reference to mutable static
   |
   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: mutable references to mutable statics are dangerous; it's undefined behavior if any other pointer to the static is used or if any other reference is created for the static while the mutable reference lives
```

With edition 2024, the lint will probably default to deny. (https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html)

To avoid that, set #[allow(static_mut_refs)] locally in the macro expansion.

This only silences the warning and doesn't answer the underlying question if we want to do that transform at all. See eg. https://github.com/rust-embedded/cortex-m/issues/411 for discussion.

### DIFF
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -103,6 +103,7 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
         #[doc(hidden)]
         #[export_name = "main"]
         pub unsafe extern "C" fn #tramp_ident() {
+            #[allow(static_mut_refs)]
             #ident(
                 #(#resource_args),*
             )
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks. I think it makes sense to allow the lint for now, since it's otherwise hard for the user to do anything about.

adamgreig (approved):

PR #562, opened 2024-10-18
Fix CI after update of ubuntu and rust

jannic:
The qemu package is no longer available. Fortunately, it's not really used by the CI script. It uses qemu-system-arm, which is also installed.

(See https://github.com/rust-embedded/cortex-m/actions/runs/11363980750/job/31609061545?pr=560 as an example for a pipeline failing without this fix.)

And nightly introduced some new lints that broke the build due to `--deny warnings`.

### DIFF
--- a/.github/workflows/on-target.yml
+++ b/.github/workflows/on-target.yml
@@ -23,7 +23,7 @@ jobs:
           RUSTFLAGS: -C link-arg=-Tlink.x -D warnings
         run: cargo build -p testsuite --target thumbv7m-none-eabi
       - name: Install QEMU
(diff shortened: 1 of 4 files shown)


jannic:
Well. That fixes one error. Unfortunately rust had a new release as well. CI hell!

jannic:
The remaining failing CI check is already covered by 
https://github.com/rust-embedded/cortex-m/pull/561

jannic:
As I understand the CI pipeline, the check fixed in #561 is not marked as required and is therefore should not block merging #562.
If that doesn't work, I can of course merge the two fixes into a single pull request.

adamgreig (approved):

PR #563, opened 2024-10-21
Fix wording on clear_current

Abrahamh08:
The current wording implies the next call to `has_wrapped()` after `clear_current()` always returns false. According to the arm docs this is not always true:
image


adamgreig (approved):

adamgreig:
thanks!

PR #565, opened 2024-10-30
cortex-m-rt: Added `_stack_end` symbol for use with MSPLIM

### LINKED ISSUE #564: `cortex-m-rt` linker script symbols are hard to use with `thumbv8` MSPLIM
Hi all,

I'm adding `flip-link`-like functionality to a project using a Cortex-M33, so I plan to set MSPLIM to the end of statically
allocated RAM. I checked the `cortex-m-rt` crate for available public symbols to set MSPLIM, and I want to set it to `__euninit`, which is where the stack and statically allocated memory collide. However, there isn't a public symbol for this. Moreover, double-underscore symbols are private as stated here:

https://github.com/rust-embedded/cortex-m/blob/6a324a830fd3040181d18ce59ae05bd44e4759a6/cortex-m-rt/link.x.in#L3-L7

What would be the best course of action here?

To me it seems that either one adds a feature like `set-msplim` to `cortex-m` to extend `cortex_m::register::msplim` with a method that uses the internal symbol from `cortex-m-rt`, or add a public symbol in `cortex-m-rt` to indicate "end of statically allocated memory" so users can setup MSPLIM themselves in `pre_init`.

Right now I'll do the following, but it is technically allowed to break in any patch release (even if we don't do that in practice):

```rust
#[cortex_m_rt::pre_init]
unsafe fn pre_init() {
    extern "C" {
        static __euninit: u32;
    }

    msplim::write(addr_of!(__euninit) as u32);
}
```

korken89:
I have added the `_stack_end` as discussed in the meeting.

What I'm not 100% sure on is how this interacts with `__sheap`, as I can't really find documentation for what its guarantees are - plus that the symbol is private so no user code and safely use it.

Closes #564

### DIFF
--- a/cortex-m-rt/CHANGELOG.md
+++ b/cortex-m-rt/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 - MSRV is now Rust 1.61 to support syn verions >=2.0.68
+- The `_stack_end` symbol is added, e.g. for use with MSPLIM.
 
 ## [v0.7.4]
 
--- a/cortex-m-rt/link.x.in
+++ b/cortex-m-rt/link.x.in
@@ -185,6 +185,9 @@ SECTIONS
   /* Place the heap right after `.uninit` in RAM */
   PROVIDE(__sheap = __euninit);
 
+  /* Place stack end at the end of allocated RAM */
+  PROVIDE(_stack_end = __euninit);
+
   /* ## .got */
   /* Dynamic relocations are unsupported. This section is only used to detect relocatable code in
      the input files and raise an error if relocatable code is found */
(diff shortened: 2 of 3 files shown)


adamgreig:
Thanks! I think lining up with `__sheap` should be fine; by default that's what you'd want MSPLIM etc to be set to anyway.

My only outstanding concern is that the new docs say users must specify `_stack_end` if they set `_stack_start`, but a lot of people are already setting `stack_start` and this isn't meant to be a breaking change. In particular it ended up in a lot of templates and examples so has been copied to `memory.x` far and wide, annoyingly usually with `_stack_start = ORIGIN(RAM) + LENGTH(RAM);` which is the default anyway! Perhaps the documentation around setting `_stack_start` could be clearer about when you must also set `_stack_end` and what will happen if you don't?

korken89:
Thanks for the comment @adamgreig, I've updated the wording to be more clear.

adamgreig (approved):

PR #573, opened 2025-01-11
cortex-m-rt: fix off-by-one in .text in FLASH section assert

9names:
The original assertion states that is checking that .text is inside flash, but it also precludes .text being at the start of flash.

This shouldn't affect anyone using the linker script from within cortex-m-rt as the vector table is always at the start. Only folks using this script as a template for a custom linker script would notice this issue.

Background:
I was using this linker script as a template for writing a run-from-ram binary for rp2040.
The way the rp2040 tools identify a run-from-ram binary is that the entry point is the first thing in the binary.
So I moved the vector table back, and the Entry function to the start, and then hit this assertion.


adamgreig (approved):

adamgreig:
thanks!

PR #577, opened 2025-02-08
Fix _stack_end usage and doc

tlevora:
1) `paint-stack` feature used `__sheap` as a lower boundary. It is not an error
  until the `_stack_end` is overrided by `cortex-m-rt` user. If overrided, it
  can potentially cause writting over memory between `__sheap` and
  `_stack_end` addresses.

2) `_stack_end` is checked for being less or equal to `_stack_start`, but
  documentation declared `_stack_end` is not used internally.


adamgreig (approved):
Thanks, looks good!

PR #578, opened 2025-02-09
Fix stack end placement

tlevora:
The stack end / heap start is placed just after `.uninit` section. It is fine until a custom section is placed just after the `.uninit` section. If application places a custom section just after the `.uninit` section, there is a collision of the custom section and heap start / stack end.

There is still a solution since both `__sheap` and `_stack_end` are designed to be overrided by an application, but changing their placement in the proposed way can save some debugging time. Proposed fix does not change behavior if no custom section added.

### DIFF
--- a/cortex-m-rt/link.x.in
+++ b/cortex-m-rt/link.x.in
@@ -182,11 +182,12 @@ SECTIONS
     __euninit = .;
   } > RAM
 
-  /* Place the heap right after `.uninit` in RAM */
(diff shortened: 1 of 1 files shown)


jonathanpallant (approved):

jonathanpallant:
Looks good to me. @adamgreig ?

adamgreig:
Yep, LGTM. Thanks @tlevora.

PR #580, opened 2025-02-12
cortex-m-rt: Add optional MSPLIM initialization, fix vector table size on armv8m

decaday:
This PR adds an optional initialization for the Main Stack Pointer Limit (MSPLIM) register, controlled by the new `set-msplim` feature and `armv8_main` cfg.

The `Main Stack Pointer Limit (MSPLIM)` register is used to define the lowest valid stack address, helping to enforce stack limits and prevent unintended stack overflows. This register is only available on ARMv8-M Mainline (Cortex-M33).  

MSPLIM initialization is often handled in the startup assembly file:

https://github.com/ARMmbed/uvisor/blob/97906335ea6892866e1fab7c6ef287004233826f/core/system/src/main.c#L84-L86

Some chip bootloaders also configure MSPLIM in advance. If it is not explicitly initialized in the startup code of user firmware, it will cause a HardFault when the program pushes to the stack in main().

Instead of relying on HALs or applications to define `__pre_init` for `MSPLIM` initialization, it seems more appropriate to handle this directly in cortex-m-rt, I think.

### DIFF
--- a/cortex-m-rt/CHANGELOG.md
+++ b/cortex-m-rt/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add `set_msplim` feature to conditionally set the MSPLIM register at device
+  reset ([#580]).
+
 ## [v0.7.5]
 
 - Fix incorrect dependency on cortex-m-rt-macros in v0.7.4 which led to
--- a/cortex-m-rt/Cargo.toml
+++ b/cortex-m-rt/Cargo.toml
@@ -45,6 +45,7 @@ required-features = ["device"]
 device = []
 set-sp = []
 set-vtor = []
+set-msplim = []
 zero-init-ram = []
 paint-stack = []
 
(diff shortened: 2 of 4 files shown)


decaday:
Fixed interrupt count for Cortex-M23 (240) and Cortex-M33 (480) based on the technical reference manual. The original value of 496 referred to the entire vector table length, with exceptions length already included in the assert check.

adamgreig (approved):
Thanks, this looks good!

jannic:
Works as expected on an rp2350.
Tested using rp235x-hal, with a trivial example using unbounded recursion.
I get a HardFault in with and without msplim, but with msplim it happens right when the stack limit is reached:
```
(gdb) print $sp
$2 = (*mut ()) 0x20000450
```
While without msplim, it happens only when running out of RAM (after overwriting all the heap):
```
(gdb) print $sp
$1 = (*mut ()) 0x1fffffd0
```
After enabling SHCSR.USGFAULTENA I also see the UsageFault being called, and the STKOF bit in UFSR being set. (*0xe000ed28 == 0x100000)

PR #582, opened 2025-03-06
cortex-m-rt: Add alignment check for the `.vector_table`

korken89:
This adds an alignment check to the placement of the vector table in accordance to armv6-m, armv7-m, and armv8-m requirements so if a user does something wrong with the linker scripts `cortex-m-rt` will throw an error.


adamgreig (approved):

adamgreig:
Thanks!

Issue #151, opened 2019-07-01
Delay time influenced by memory configuration


tpwrules:
How many cycles `asm::delay()` can be affected by how the chip memory is configured and even where the function ends up in Flash. The code assumes the loop takes 4 cycles and explains that interrupts may cause the function to delay longer than requested.

However, items such as Flash wait states, instruction cache, and function alignment can also cause the loop to take longer, even when interrupts are disabled. As an example, `asm::delay()` running on the NRF52840 can take 4, 5, or 6 cycles per loop depending on whether the function ends up 16, 8, or 2 byte aligned. This also means that the delay depends on how exactly the binary is compiled and linked.

This is very surprising behavior and I think a note to the above effect would be useful in the function's documentation. Perhaps the loop can be measured as part of initialization, but this would probably be unnecessary overhead.

therealprof:
The documentation clearly states that delay will block for **at least** the specified amount of cycles. It certainly is a really poor delay implementation and not meant to be used for accurate timing. PRs to improve upon that would certainly be appreciated.

NB: Since this function is inlined it will behave differently depending on the caller so measuring as part of the initialisation is not possible. Not inlining it would make it even less dependable, at least for short delays.

ianrrees:
I believe the concerns in the OP here have been addressed (to the extent they can be, as discussed in other issues) now that #587 has been merged.

Though, that is a curious note about the NRF52840 and 16 byte alignment.  I only tested #587 on a SAMD51 (Cortex M4) on which the branch timing seems to only differentiate between 2, 4, or 8 byte aligned targets.

tpwrules:
I don’t have the bandwidth right now to confirm the fix or sort out why I came to the cycle vs. alignment conclusions that I did, but this looks like a good step to me.

Thanks for working on it. Feel free to close the issue.

ianrrees:
I agree about the "at least", will aim to PR a change to that wording and fix a minor error I made in the comment that explains the alignment.  In another thread, I suggested that we put the argument in terms of instructions/NOPs instead of cycles - maybe that would better communicate the situation.

With a slightly wider view, I'm wondering if the right place for this method is actually in the HAL layer instead.  Just thinking about the ATSAM one that I am most familiar with, we could quite easily make a delay that's more-or-less cycle correct since we know exactly what chip the code will run on.

adamgreig:
[...]

PR #587, opened 2025-04-14
Fix asm::delay() variability due to alignment

ianrrees:
While testing a new clocking API in [atsamd-rs](https://github.com/atsamd-rs/atsamd), we encountered a very strange behaviour which took some time to track down - seemingly innocuous changes (eg, enabling features that weren't used by the firmware, or adding a debug call) would result in wildly different execution speed of a test program.  In the context of testing at a new clocking API, at least two of us assumed the issue was down to the clock speed changing, but in fact it was the number of cycles of delay we were getting for calls to `asm::delay()`.  This turns out to be a result of the Cortex-M4 taking a variable number of cycles to execute a conditional branch, based on the alignment of the target.

Although the method is strictly conforming to the documentation by delaying /at least/ N cycles, the sensitivity to alignment is quite surprising.  This PR aligns the loop target, to remove this surprising behaviour.

Resolves part of https://github.com/rust-embedded/cortex-m/issues/151

### DIFF
--- a/cortex-m/src/asm.rs
+++ b/cortex-m/src/asm.rs
@@ -33,6 +33,10 @@ pub fn delay(cycles: u32) {
     let real_cycles = 1 + cycles / 2;
     unsafe {
         asm!(
+            // The `bne` on some cores (eg Cortex-M4) will take a different number of instructions
+            // depending on the alignment of the branch target.  Set the alignment of the top of the
+            // loop to prevent surprising timing changes when the alignment of the delay() changes.
+            ".p2align 3",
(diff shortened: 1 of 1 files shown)


ianrrees:
I suppose once [fn_align](https://github.com/rust-lang/rust/issues/82232) is stable, that could be another option although it probably won't make much difference in practice.

adamgreig (approved):

adamgreig:
Thanks for investigating and for this improvement, I think it's well worth the tradeoff to get better determinism.

PR #591, opened 2025-04-16
cortex-m-semihosting: make semihosting macros compatible with 1-arg format strings

### LINKED ISSUE #581: `cortex-m-semihosting`: `hprint!` family of macros do not format strings without arguments
In the `hprint!` macros:

https://github.com/rust-embedded/cortex-m/blob/f2fc7d3ee13cc23e2b5e7d05d393ad8257197a26/cortex-m-semihosting/src/macros.rs#L36-L44

They do not print format strings with no arguments (e.g., `hprint!("{address}")` is sugar for `hprint!("{}", address)`, which is a deviation from the stdlib `print!` macros using `format_args!`.  

If it's acceptable performance-wise, I can submit a PR to use e.g. `hstdout_fmt` in this case instead of `hstdout_str`, otherwise I can document this deviation.

suaviloquence:
Fixes #581

### DIFF
--- a/cortex-m-semihosting/src/macros.rs
+++ b/cortex-m-semihosting/src/macros.rs
@@ -36,7 +36,7 @@ macro_rules! syscall1 {
 #[macro_export]
 macro_rules! hprint {
     ($s:expr) => {
-        $crate::export::hstdout_str($s)
+        $crate::export::hstdout_fmt(format_args!($s))
     };
     ($($tt:tt)*) => {
         $crate::export::hstdout_fmt(format_args!($($tt)*))
@@ -53,10 +53,10 @@ macro_rules! hprintln {
         $crate::export::hstdout_str("\n")
     };
     ($s:expr) => {
-        $crate::export::hstdout_str(concat!($s, "\n"))
+        $crate::export::hstdout_fmt(format_args!("{}\n", format_args!($s)))
     };
     ($s:expr, $($tt:tt)*) => {
-        $crate::export::hstdout_fmt(format_args!(concat!($s, "\n"), $($tt)*))
+        $crate::export::hstdout_fmt(format_args!("{}\n", format_args!($s, $($tt)*)))
     };
 }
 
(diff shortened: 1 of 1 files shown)


cortex-m-semihosting/src/macros.rs
```diff
@@ -84,10 +84,10 @@ macro_rules! heprintln {
         $crate::export::hstderr_str("\n")
     };
     ($s:expr) => {
-        $crate::export::hstderr_str(concat!($s, "\n"))
+        $crate::export::hstderr_fmt(format_args!("{}\n", format_args!($s)))
     };
     ($s:expr, $($tt:tt)*) => {
-        $crate::export::hstderr_fmt(format_args!(concat!($s, "\n"), $($tt)*))
+        $crate::export::hstderr_fmt(format_args!("{}\n", format_args!($s, $($tt)*)))
```
jonathanpallant, line 90: Is the double nested `format_args!` required here? Was there some use case that didn't work with the previous method of using `concat!`?

jonathanpallant:
As a point of reference, the [semihosting](https://crates.io/crates/semihosting) crate does it like this: https://docs.rs/semihosting/0.1.18/src/semihosting/macros.rs.html.

You could make the argument we should deprecate this crate and recommend people use that one.

cortex-m-semihosting/src/macros.rs
suaviloquence: Yes, sorry I meant to take a note of it - `format_args!` doesn't expand captures in the format string literal if it is generated by a macro:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=154ba94b6bc1e4428eed13cf3e1f035f)

From https://github.com/rust-lang/rust/pull/106824, it should get flattened into a single `format_args!/Arguments` call.

jonathanpallant: yikes - thanks for the clarification.

suaviloquence:
Anything else I need to do on this PR?

jonathanpallant:
No, it's all good. I was just letting others have a chance to weigh in.

PR #596, opened 2025-05-06
fix(coretex-m-rt-macros): Allow `#[unsafe(link_section = )]` for Rust 2024 edition

yutannihilation:
Rust 2024 edition requires marking `link_section` attribute as `unsafe`. Otherwise, the compilation fails with the following error.

https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-attributes.html

```
error: unsafe attribute used without unsafe
  --> src/bin/tmp.rs:21:3
   |
21 | #[link_section = ".bi_entries"]
   |   ^^^^^^^^^^^^ usage of unsafe attribute
   |
help: wrap the attribute in `unsafe(...)`
   |
21 | #[unsafe(link_section = ".bi_entries")]
   |   +++++++                            +

error: could not compile `rp2350-blink` (bin "tmp") due to 1 previous error
```

However, if I use `#[unsafe(link_section = ".bi_entries")]`, cortex-m-crate complains that it's not allowed (cf. https://github.com/embassy-rs/embassy/pull/4170).

```
   Compiling embassy-nrf v0.3.1 (/ci/code/embassy-nrf)
error: this attribute is not allowed on an interrupt handler controlled by cortex-m-rt
  --> /ci/code/embassy-rp/src/multicore.rs:93:1
   |
93 | #[unsafe(link_section = ".data.ram_func")]
   | ^

error: this attribute is not allowed on an interrupt handler controlled by cortex-m-rt
   --> /ci/code/embassy-rp/src/multicore.rs:118:1
    |
118 | #[unsafe(link_section = ".data.ram_func")]
    | ^

   Compiling embassy-stm32 v0.2.0 (/ci/code/embassy-stm32)
```

This pull request tweaks the check logic to allow this notation so that users can use Rust 2024 edition.

### DIFF
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -755,8 +755,8 @@ fn check_attr_whitelist(attrs: &[Attribute], caller: WhiteListCaller) -> Result<
     ];
 
     'o: for attr in attrs {
-        for val in whitelist {
-            if eq(attr, val) {
+        if let Some(attr_name) = get_attr_name(attr) {
+            if whitelist.contains(&attr_name.as_str()) {
                 continue 'o;
             }
         }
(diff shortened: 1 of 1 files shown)


BartMassey:
Looks good to me.

therealprof:
How does this PR figure out whether edition 2024 is in play or not? Or is that up to the user to figure out? It would be nice if there was a way to provide guidance about that change between editions.

yutannihilation:
`#[unsafe(link_section = )]` is a valid syntax both in edition 2021 (Rust >=1.85) and in edition 2024. The difference is that it's mandatory in edition 2024. The user will immediately see the guidance in the error message by the Rust compiler if they tries to compile without `unsafe()`. So, I believe there's nothing that has to be done on cortex-m-rt-macros' side. Any concerns...?

therealprof:
> `#[unsafe(link_section = )]` is a valid syntax both in edition 2021 (Rust >=1.85) and in edition 2024. The difference is that it's mandatory in edition 2024. The user will immediately see the guidance in the error message by the Rust compiler if they tries to compile without `unsafe()`. So, I believe there's nothing that has to be done on cortex-m-rt-macros' side. Any concerns...?

Sounds excellent to me then. No concerns. 😄

therealprof (approved):

yutannihilation:
Thanks for reviewing!

Sympatron:
> is a valid syntax both in edition 2021 (Rust >=1.85)

The docs currently state an MSRV of 1.65: https://github.com/rust-embedded/cortex-m/blob/c04e4db3caf2e2ca38fd486f66b9a6176fe3f10c/cortex-m/src/lib.rs#L35-L36

therealprof:
> The docs currently state an MSRV of 1.61:

This doesn't affect MSRV. People can still use the old "safe" `link_section` attribute. This only changes that you **can** use the `unsafe link_section` attribute at all. Starting with Rust 1.85 you can use either, with edition 2024 it's a **MUST** use.

PR #598, opened 2025-05-16
Mark `pre_init` as deprecated

### LINKED ISSUE #595: Deprecate `#[cortex_m_rt::pre_init]`?
The docs say that using `cortex_m_rt::pre_init` is deprecated, but the macro is not effectively deprecated. Maybe we should add something like:

```
#[deprecated(note = "Use global_asm! to define the __pre_init function instead")]
```

romancardenas:
Closes #595


therealprof (approved):
LGTM

PR #601, opened 2025-06-11
cortex-m-rt: Allow custom memory region names

kpfleming:
Adds support for environment variables which allow the user to use their own memory region names in memory.x, instead of FLASH and/or RAM.


kpfleming:
I don't understand why `far` had to be added to *both* `dev-dependencies` and `build-dependencies` in `Cargo.toml`, but I couldn't get a successful build without it.

jonathanpallant:
Two thoughts:

1. What does `far` do that [`replace`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.replace) does not?
2. What's the motivation for wanting to be able to have different region names in the linker script, and can people just use region aliases to achieve the same outcome?

kpfleming:
Closing this PR, as region aliases are an easy solution. I'll contribute an example to the embassy-boot docs :-)

Issue #609, opened 2025-09-17
Add feature to disable .data copy when using a copying bootloader


BenChung:
This is an issue that I found when getting boot2's copy-to-RAM setup working on RP2040. The boot2 bootloader sets up the SPI flash, copies all of it (including the .data region) into memory, and then de-initializes the SPI flash so it's no longer mapped into the memory space. If the LMA and VMAs are set up correctly with `__sidata = LOADADDR(.data)` cortex-m-rt's apploader will then come along and try to copy the `.data` region out of flash again, but since the SPI flash is no longer mapped it gets nonsense bytes.

I was able to proximally solve the issue by setting `__sidata = ADDR(.data)` in my linker script, which causes the app loader to copy the `.data` region to itself in RAM after boot2 is done. A better solution would be to add a feature flag to the library that lets me disable the copy entirely (which would then let me use a correct definition for `__sidata`), but I wanted to open an issue to see if there's a better way to solve it before opening a PR.

9names:
@Sympatron the context is in the first sentence. you weren't doing the same thing as OP.

> when getting boot2's copy-to-RAM setup

https://github.com/rp-rs/rp2040-boot2/tree/06402359ca3ec2ef8ce3b8dbade7dc126f95781c#booting-from-ram

https://github.com/rp-rs/rp2040-boot2/blob/06402359ca3ec2ef8ce3b8dbade7dc126f95781c/src/boot2_ram_memcpy.S#L57

Sympatron:
@9names: You are right, I looked it up. I was using `rp2040_boot2` indirectly through `rp-pico` which uses `rp2040_boot2::BOOT_LOADER_W25Q080`. That is why this was never a problem for me.

BenChung:
Yes, by default rp-pico uses XIP while running which keeps the RAM mapped, so the boot script keeps working.

This project https://github.com/BenChung/RP2040_boot2_run_from_ram should be set up to run from RAM, though the Cargo.toml needs to be updated to point at your local copy of embassy and defmtt. It has the fix applied in the `link.ram.x` linker script:

```
  /* LMA of .data */
  /* should actually be __sidata = LOADADDR(.data) - but by setting it to ADDR(.data) it works */
  __sidata = ADDR(.data);
```

Replace this line with 
```
  __sidata = LOADADDR(.data);
```
to reproduce the problem (tested on a RP2040 in a pico).

PR #613, opened 2025-10-06
Add skip-data-init feature to disable .data section initialization

KushalMeghani1644:
## Introduces:
- the skip-data-init feature that disables the .data copy loop in the reset handler, for boot flows that already initialize RAM (e.g. RP2040 boot2 that copies to RAM and unmaps flash).

- Adds a compile-time guard preventing skip-data-init from being used with zero-init-ram.

- Includes a minimal example at [skip-data-init.rs](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

- Default behavior is unchanged; .data is still copied when the feature is not enabled.

### This PR tried to address issue #609

### DIFF
--- a/cortex-m-rt/CHANGELOG.md
+++ b/cortex-m-rt/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Mark `pre_init` as deprecated
 - Add `set_msplim` feature to conditionally set the MSPLIM register at device
   reset ([#580]).
+- Add `skip-data-init` feature to optionally skip copying the `.data` section.
 
 ## [v0.7.5]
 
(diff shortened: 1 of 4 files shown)


jonathanpallant:
Stubbing out the copy loop seems fine but what is the purpose of the example? It has no observable output and no mechanism to run it with the kind of bootloader that it must be used with to avoid UB. Was it LLM generated?

KushalMeghani1644:
Hey @jonathanpallant the purpose of that example is to demonstrate how a Cortex-M program behaves when the usual `.data` section initialization is skipped by cortex-m-rt. This is useful because normally the cortex-m-rt copies `.data` section from flash to RAM. But for some scenarios when using the RP2040 boot2 bootloader as stated in issue #609 this copy has been already done by the bootloader itself, and the cortex-m-rt copies again, basically it reads from unmapped memory causing undefined behaviour.

Either way my PR isn't LLM generated, through that example I was indeed trying to show something reasonable :)

Thanks!

jonathanpallant:
Ok so how do I run the example and what will I observe when the feature is set?

PR #615, opened 2025-10-07
Add skip-data-copy feature for when the .data LMA isn't initialized after boot.

### LINKED ISSUE #609: Add feature to disable .data copy when using a copying bootloader

BenChung:
This implements the `skip-data-copy` feature, which disables loading the `.data` section (containing statically allocated variables) in cases when that section is either:

* Already loaded and possible, or
* No longer valid when the startup code runs.

This occurs for example with boot2 using the `BOOT_LOADER_RAM_MEMCPY` option, which copies the `.data` section and then de-initializes XIP before relinquishing control leaving the contents of `.data` at the LMA undefined.

This fixes #609. Tested on a RP2040 in a pico. The test project is https://github.com/BenChung/RP2040_boot2_run_from_ram, which panics without this fix, or works correctly with this patch (shown on the branch https://github.com/BenChung/RP2040_boot2_run_from_ram/tree/fixed).


jannic:
And successfully tested using rp2040-hal. Good job!

PR #630, opened 2026-03-13
Run clippy against Arm build targets.

wt:
The current clippy runs are running against an X86 build of the code This will make clippy exercise the Arm specfic code. I have only enabled 2 as some jobs were cancelled when I enabled all 7 targets.


thejpster (approved):
LGTM, but I think we need to change the name of the job that PRs require before merging, because this matrix changes the name of the jobs?

Pinging @adamgreig for a second opinion.

thejpster:
A rebase should fix this?

wt:
I think so.

PR #632, opened 2026-03-22
Extend the test suite to check more of our assembly code

thejpster:
We weren't testing our `#[inline]` asm code thanks to https://github.com/rust-lang/rust/issues/147275. So this PR extends the testsuite to run a few of them. This will at least check they have valid syntax.

### DIFF
--- a/cortex-m/src/psp.rs
+++ b/cortex-m/src/psp.rs
@@ -79,29 +79,48 @@ impl<const N: usize> core::default::Default for Stack<N> {
 /// In Unprivileged Mode, code can no longer perform privileged operations,
 /// such as disabling interrupts.
 ///
-#[cfg(cortex_m)]
-pub fn switch_to_unprivileged_psp(mut psp_stack: StackHandle, function: extern "C" fn() -> !) -> ! {
-    // set the stack limit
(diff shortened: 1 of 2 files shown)


cortex-m/src/psp.rs
```diff
@@ -79,29 +79,48 @@ impl<const N: usize> core::default::Default for Stack<N> {
 /// In Unprivileged Mode, code can no longer perform privileged operations,
 /// such as disabling interrupts.
 ///
-#[cfg(cortex_m)]
-pub fn switch_to_unprivileged_psp(mut psp_stack: StackHandle, function: extern "C" fn() -> !) -> ! {
-    // set the stack limit
-    #[cfg(armv8m_main)]
-    unsafe {
-        crate::register::psplim::write(psp_stack.bottom() as u32);
+pub fn switch_to_unprivileged_psp(psp_stack: StackHandle, function: extern "C" fn() -> !) -> ! {
```
jannic: Doesn't this need to be `unsafe`? After all, it calls FFI code.

(Of course, this is not new with this PR.)

jannic (approved):

PR #641, opened 2026-04-10
use const steal

xoviat:
The steal function is already unsafe, so there's no need to mark taken to make sure that someone else doesn't take it.

### DIFF
--- a/cortex-m/src/peripheral/mod.rs
+++ b/cortex-m/src/peripheral/mod.rs
@@ -169,6 +169,7 @@ impl Peripherals {
     }
 
     /// Steal the peripherals without marking them as taken.
+    #[inline]
     pub const unsafe fn const_steal() -> Self {
         Peripherals {
             #[cfg(feature = "cm7")]
--- a/cortex-m/src/peripheral/mod.rs
+++ b/cortex-m/src/peripheral/mod.rs
@@ -168,60 +168,65 @@ impl Peripherals {
         })
     }
 
+    /// Steal the peripherals without marking them as taken.
+    pub const unsafe fn const_steal() -> Self {
+        Peripherals {
+            #[cfg(feature = "cm7")]
+            AC: AC {
+                _marker: PhantomData,
+            },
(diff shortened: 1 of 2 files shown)


wt:
Doesn't the `steal` function need to mark the peripherals taken so that `take` will work as intended? Otherwise, using `take` after `steal` will get the peripherals when we don't know that there is exclusive access via take.

xoviat:
> Otherwise, using take after steal will get the peripherals when we don't know that there is exclusive access via take.

Yes, and this is the correct result. `steal` is already unsafe, therefore, if you call it, you cannot guarantee that other code somewhere else does not have a duplicate instance of the peripherals. It is therefore your job to make sure that this isn't a problem.

However, suppose you are writing unsafe code and you need to call cortex-m methods, while still allowing safe code to take them. There is no way to do so without using a transmute, because calling `steal` will impact `take`.

adamgreig:
This is an annoying legacy thing from the early ecosystem days when cortex-m, svd2rust, and cortex-m-rtic (then -rtfm) were a bit more co-designed, but [at the moment RTIC](https://github.com/rtic-rs/rtic/blob/ae9b5304ade0b225b440b4e3e9b156912fe28a03/rtic-macros/src/codegen/pre_init.rs#L17) (and perhaps other things we don't know about) uses `steal()` and requires that it marks the peripherals as taken, so that future calls to `take()` fail. In other words, it's required that `steal()` impacts `take()` and other libraries rely on that for correctness.

I think the only option (short of a breaking change in 0.8 or 1.0 where we change the API completely) is to add a new method that is `const unsafe` and is documented to not affect `take()`. I can't find it right now, but there was a lot of back and forth on this topic a couple of years ago, mostly trying to bikeshed a suitable name for it.

xoviat:
@adamgreig What about 'const_steal'?

jonathanpallant:
What use-case is this intended to support, that the current APIs do not support?

I'd be wary about adding this given how the existing `steal` function behaves.

Issue #644, opened 2026-04-11
Make it more obvious that the `inline-asm` feature is deprecated and does nothing


wt:
The `inline-asm` feature is a relic of the time before inline asm was supported in stable rust. We should do something to make it more obvious that that feature now does nothing and will be removed in a future major version.

One idea might be to at least have a message from build.rs that is output when the `inline-asm` feature is active is used.

PR #645, opened 2026-04-11
Document deprecation of inline-asm feature

9names:
Add deprecation notice for `inline-asm` to readme and feature declaration in Cargo.toml.
Remove documentation for deprecated feature `inline-asm` in the rest of this crate.

Reasonable first steps for #644


wt:
This all looks great. I would approve if I could.

adamgreig (approved):

PR #656, opened 2026-04-27
Remove bare-metal dependency

robamu:
Includes #654


djc:
@adamgreig friendly ping?

adamgreig:
This is a breaking change, so I don't think it will be merged until after we release 0.7.8 and start the 0.8 window.

PR #657, opened 2026-04-28
update and improve testsuite

robamu:
- Use `defmt-test` in testsuite
- Add tests for Cortex-M0 as well


therealprof (approved):
Nice. I like it.

PR #664, opened 2026-06-20
fix: read ICSR volatile in SCB::vect_active()

jakobjung10:
`SCB::vect_active()` now reads ICSR with a volatile access, preventing the load from being narrowed to a byte read that faults on word-only ICSR models.

### DIFF
--- a/cortex-m/CHANGELOG.md
+++ b/cortex-m/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - The code should be clippy clean on all supported configurations now.
 - Add embedded-hal 1.0 delays.
 
+### Fixed
+- `SCB::vect_active()` now reads ICSR with a volatile access, preventing the load from being narrowed
+  to a byte read that faults on word-only ICSR models.
+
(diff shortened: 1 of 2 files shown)


jonathanpallant:
That's a plain read not a volatile read?

jakobjung10:
`read()` in ```(*SCB::PTR).icsr.read()``` internally calls [self.register.get()](https://github.com/rust-embedded/volatile-register/blob/master/src/lib.rs#L77) from [volatile-register](https://github.com/rust-embedded/volatile-register/tree/master), which again internally calls [ptr::read_volatile(self.value.get())](https://github.com/japaric/vcell/blob/master/src/lib.rs#L33) from [vcell](https://github.com/japaric/vcell/tree/master).
Therefore, it is a volatile read.

Also, calling `read()` aligns more with the remaining scb module, as the direct call of `ptr::read(&(*SCB::PTR).icsr as *const _ as *const u32)` is actually the outlier.

jonathanpallant:
oof, fair point - my mistake!

jonathanpallant (approved):

adamgreig (approved):

PR #665, opened 2026-06-23
Add note about change of asm::delay behaviour to changelog

adamgreig:


thejpster (approved):

jonathanpallant (approved):
OK, fine, if I wear this hat I have more permissions

PR #666, opened 2026-07-10
Improve TrustZone support

jonathanpallant:
Improves TrustZone support, for anyone writing a secure mode application.

- Add secure-mode feature, to enable access to SCB_NS peripheral
- Allow access to SFAR and SFSR fields

With this I can write a useful handler for SecureFault exceptions, and I can
set the VTOR for nonsecure mode from my secure mode bootloader.

PR #671, opened 2026-07-21
Improve trustzone support: take2

jonathanpallant:
This PR merges #648 and #666 into one cohesive whole.


cortex-m/src/asm.rs
```diff
@@ -390,6 +390,80 @@ pub unsafe fn bootload(vector_table: *const u32) -> ! {
...
+        mov     r11, 0
+        mov     r12, 0
+        // r13 = MSP, which we set earlier
+        // r14 = LR, which is the address we jump to
+        // r15 = PC, which bxns will set
+        bxns    lr
+        .size _bx_ns_trampoline, . - _bx_ns_trampoline
+    "#,
+);
+
```
diondokter, line 466: This screams naked function. But I'm guessing the MSRV is too low for that?
Anyways this should work.

The only thing I remember that the normal calling conventions will also clear the processor flags.
See the msr instruction in the example assembly here: https://github.com/rust-lang/rust/issues/81391

jonathanpallant: Good idea - done

jonathanpallant:
I have tested this bootloader function on an STM32U5:

```text
Hello, this is secure-loader. Configuring peripherals...
...SAU configured
...GTZC1 configured
Booting NS application at 0x08200000...
SECURE FAULT:
- SFAR = 20000000
- Attribution Unit Violation
```

```rust
#[cortex_m_rt::entry]
fn main() -> ! {
    let mut bsp: bsp::SecureBoard = bsp::SecureBoard::new();

    // Enable secure fault handler
    bsp.scb
        .enable(cortex_m::peripheral::scb::Exception::SecureFault);

    // Say hello
    println!("Hello, this is secure-loader. Configuring peripherals...");

    // Configure peripherals
    bsp.configure_sau();
    println!("...SAU configured");
    bsp.set_sram3_nonsecure();
    println!("...GTZC1 configured");

    let ns_app_base = bsp::hal::ns_addr::FLASH2_START;
    println!("Booting NS application at 0x{:08x}...", ns_app_base);
    // Boot nonsecure world
    unsafe {
        cortex_m::asm::bootload_ns(ns_app_base as *const u32, bsp.scb_ns);
    }
}
```

jonathanpallant:
`VLSTM` is apparently a NOP if you don't have an FPU, so maybe we should unconditionally do it? Avoids issues if you were using the FPU despite running a soft-float (EABI) target.

See https://developer.arm.com/documentation/dui0801/l/Floating-point-Instructions--32-bit-/VLSTM--A32-

diondokter:
That sounds good. Keeps complexity low and it always does the correct thing. The only cost is 2 bytes of flash and 1 clock cycle. Easily worth it

jonathanpallant:
These changes also allow a SecureFault handler that looks like this:

```rust
#[cortex_m_rt::exception]
fn SecureFault() {
    println!("SECURE FAULT:");
    // Safety: No-one else is using the SAU, so this won't race, plus the registers we want are read-only
    let sau = unsafe { &*cortex_m::peripheral::SAU::PTR };
    let sfsr = sau.sfsr.read();
    if sfsr.sfarvalid() {
        println!("- SFAR = {:08x}", sau.sfar.read().address());
    }
    if sfsr.invep() {
        println!("- Invalid Entry Point");
    }
    if sfsr.invis() {
        println!("- Invalid Integrity Signature");
    }
    if sfsr.inver() {
        println!("- Invalid Exception Return");
    }
    if sfsr.auviol() {
        println!("- Attribution Unit Violation");
    }
    if sfsr.invtran() {
        println!("- Invalid Transition");
    }
    if sfsr.lsperr() {
        println!("- Lazy state preservation error");
    }
    if sfsr.lserr() {
        println!("- Lazy state error");
    }
    loop {}
}
```

therealprof (approved):
LGTM!

PR #674, opened 2026-07-24
secure routing API for NVIC now takes exclusive reference

robamu:
these methods do a non-atomic operation because they modify the register. In the `set_priority` method which also does a register modification, the method takes exclusive reference by taking &mut self. maybe we should do the same here? if not, we should either update the safety note, or update `set_priority` to be consistent.


thejpster:
Yes these methods should take &mut - good catch

thejpster (approved):

jonathanpallant (approved):

PR #672, opened 2026-07-21
Release cortex-m

jonathanpallant:
* Updates cortex-m and cortex-m-macros to 0.7.8
* ~~Updates cortex-m-rt and cortex-m-rt macros to 0.7.6~~ (pulled into #675)
* Added relevant headers to both CHANGELOGs
* Fixed a bunch of doc issues that CI missed (`cargo doc --target=thumbv8m.main-none-eabihf` is now clean)

PR #676, opened 2026-07-31
remove duplicate r8 zeroing

robamu:


jonathanpallant:
I added this fix to #672 but I guess it doesn't hurt to do it separately

jonathanpallant (approved):

PR #679, opened 2026-08-05
refactor(macros): bump `syn` to v3

AudaciousAxiom:
Following the [release of a new major version of `syn`](https://github.com/dtolnay/syn/releases/tag/3.0.0), this bumps the version used by `cortex-m-macros` and `cortex-m-rt-macros`, with the goal of eventually removing the duplicated dependency in binaries (and therefore reducing compile times).

`syn` v3 advertises an [MSRV of 1.71](https://crates.io/crates/syn/3.0.0/code/Cargo.toml#L14), so this shouldn't affect the MSRV of these crates given the recent bump to 1.85.

I've gone through the changelog and did the necessary refactor; I didn't find any other relevant breaking changes (but there are quite a few of them so I could have missed some). I've run `cargo test` but didn't do any other kind of testing.

### DIFF
--- a/cortex-m-rt/macros/src/lib.rs
+++ b/cortex-m-rt/macros/src/lib.rs
@@ -240,7 +240,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
         }
     };
 
-    if f.sig.unsafety.is_none() {
+    if !matches!(f.sig.safety, syn::Safety::Unsafe(_)) {
         match exn {
             Exception::DefaultHandler | Exception::HardFault(_) | Exception::NonMaskableInt => {
                 // These are unsafe to define.
@@ -635,7 +635,7 @@ pub fn pre_init(args: TokenStream, input: TokenStream) -> TokenStream {
     // check the function signature
     let valid_signature = f.sig.constness.is_none()
         && f.vis == Visibility::Inherited
-        && f.sig.unsafety.is_some()
+        && matches!(f.sig.safety, syn::Safety::Unsafe(_))
         && f.sig.abi.is_none()
         && f.sig.inputs.is_empty()
         && f.sig.generics.params.is_empty()
--- a/cortex-m/macros/Cargo.toml
+++ b/cortex-m/macros/Cargo.toml
@@ -15,7 +15,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.106"
 quote = "1.0.45"
-syn = { version = "2.0.117", features = ["extra-traits", "full"] }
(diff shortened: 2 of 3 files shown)


cortex-m-rt/macros/src/lib.rs
```diff
@@ -240,7 +240,7 @@ pub fn exception(args: TokenStream, input: TokenStream) -> TokenStream {
         }
     };
 
-    if f.sig.unsafety.is_none() {
+    if !matches!(f.sig.safety, syn::Safety::Unsafe(_)) {
```
AudaciousAxiom, line 243: Function "safety" is now encoded with [a 3-variant enum](https://docs.rs/syn/3.0.3/syn/enum.Safety.html).

AudaciousAxiom, line 243: Function "safety" is now encoded with [a 3-variant enum](https://docs.rs/syn/3.0.3/syn/enum.Safety.html).

cortex-m/macros/Cargo.toml
```diff
@@ -15,7 +15,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.106"
 quote = "1.0.45"
-syn = { version = "2.0.117", features = ["extra-traits", "full"] }
+syn = { version = ">= 2.0.117, < 4", features = ["extra-traits", "full"] }
```
AudaciousAxiom, line 18: `cortex-m-macros` seems to work with either `syn` v2 or v3: this allows both but, because `cortex-m-rt-macros` does depend on the v3 specifically, and as *most* dependents use both, you may prefer to depend on the v3 to simplify this a bit:

```suggestion
syn = { version = "3.0", features = ["extra-traits", "full"] }
```

Allowing both versions allows dependents that only depend on `cortex-m-macros` and don't otherwise have a (transitive) dependency on `syn` v3 to only use `syn` v2, thus avoiding the dependency duplication.

diondokter: I think a big pro of the simpler 3.0 dep is that it's far less likely to break something by accident. If someone accidentally uses a 3.0 feature without updating the syn dep, you're breaking everyone with syn 2 in their lock files.
Not worth it IMO

diondokter (approved):
Looks good to me, though see my thoughts on the syn version

cortex-m/macros/Cargo.toml
jonathanpallant: Agreed

AudaciousAxiom: Makes sense! I've simplified to just list `syn` v3.

therealprof (approved):
LGTM

PR #681, opened 2026-08-05
cortex-m: don't gate off modules for native target

### LINKED ISSUE #680: `cortex-m` v0.7.8 breaks crates that re-export `basepri` and `basepri_max` if those crates are compiled for non-ARM targets
In RTIC, we run (effectively) the following test:

```
cargo test -p rtic --features thumbv7-backend --target host-tuple
```

in this setup, `rtic ` uses `basepri` and `basepri_max` from `cortex-m`, i.e. it contains (effectively)

```rust
#[cfg(feature = "thumbv7-backend")]
use cortex_m::register::{basepri, basepri_max};
```

somewhere in the compiled code.

 However, in the latest release, those two modules have (correctly) been feature gated to only exist for `armv7m` and `armv8m_main`. Adam Greig mentioned that compiling for non-`cortex-m` platforms was actually a goal, so I figured I'd open an issue to indicate that this doesn't really work directly.

The fix (as far as I can tell) is to change the `#[cfg(any(armv7m, armv8m_main))]` on those modules (and perhaps others?) to `#[cfg(any(armv7m, armv8m_main, native))]`. The actually-platform-dependent items within those modules are already feature gated the same way.

We can, of course, work around this with similar `cfg`s in RTIC, but I wanted to post this here in case cortex-m would be up for fixing it. I can absolutely create a PR, this issue is mostly for discussing whether we actually want this.

datdenkikniet:
These modules only contain `asm_cfg`s that explicitly exist to allow for compilation (but not function) on native targets.

Gating the modules behind target features makes this impossible (requiring workarounds on the end of the cortex-m consumer).

The modules should also be made available when compiling for the host target.

Fixes #680 

There are additional modules that might benefit from this, but that are likely not to have compiled before (so I'm not fixing them in this PR): 

* `register::fpscr`. This has some `cfg`'d functions that need the `asm_cfg` treatment.
* `register::msplim` and `register::psplim`. Has some functions that need the `asm_cfg` treatment.
* `itm`. This one I'm having a hard time understanding. It seems like it should work, but I feel that there may be something that I'm missing, so I won't touch it for now.
* `csme`. This has correctly `asm_cfg`'d blocks, but includes a (doc?) test that starts failing if the module is enabled for `cfg(native)`.

### DIFF
--- a/cortex-m/src/register/mod.rs
+++ b/cortex-m/src/register/mod.rs
@@ -30,15 +30,15 @@
 //!
 //! - Cortex-M* Devices Generic User Guide - Section 2.1.3 Core registers
 
-#[cfg(any(armv7m, armv8m_main))]
+#[cfg(any(armv7m, armv8m_main, native))]
 pub mod basepri;
 
-#[cfg(any(armv7m, armv8m_main))]
+#[cfg(any(armv7m, armv8m_main, native))]
 pub mod basepri_max;
 
 pub mod control;
 
-#[cfg(any(armv7m, armv8m_main))]
+#[cfg(any(armv7m, armv8m_main, native))]
 pub mod faultmask;
 
 #[cfg(has_fpu)]


thejpster:
Yeah if we put in the effort to have dummy versions of the functions when running tests on the host, it doesn't make sense to remove the entire module when on the host.

thejpster (approved):

adamgreig (approved):

PR #688, opened 2026-08-20
cortex-m-rt: add reset-interrupt-state feature

rhgndf:
Some bootloaders don't disable interrupts when exiting and jumping into user code.

What should be the order to execute this with respect to set-vtor/set-sp?

### DIFF
--- a/cortex-m-rt/src/lib.rs
+++ b/cortex-m-rt/src/lib.rs
@@ -187,6 +187,14 @@
 //! conjunction with crates generated using `svd2rust`. Those peripheral access crates, or PACs,
 //! will populate the missing part of the vector table when their `"rt"` feature is enabled.
 //!
+//! ## `reset-interrupt-state`
+//!
+//! If this feature is enabled, the reset handler disables all configurable interrupts before any
+//! other startup code runs. It then stops and clears SysTick, clears pending PendSV, and disables
+//! and clears all device specific interrupts. This is useful when a bootloader jumps to the
+//! application without first disabling its interrupts. If the bootloader also leaves its vector
+//! table active, enable the `set-vtor` feature as well.
+//!
 //! ## `set-sp`
 //!
 //! If this feature is enabled, the stack pointer (SP) is initialised in the reset handler to the
@@ -556,6 +564,11 @@ cfg_global_asm! {
     ".cfi_startproc
      Reset:",
 
+    // If enabled, mask all configurable interrupts before any other startup code runs. This
+    // prevents interrupts left enabled by a bootloader from using the bootloader's vector table.
+    #[cfg(feature = "reset-interrupt-state")]
+    "cpsid i",
+
     // If enabled, initialise the SP. This is normally initialised by the CPU itself or by a
     // bootloader, but some debuggers fail to set it when resetting the target, leading to
     // stack corruptions.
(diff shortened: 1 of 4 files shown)


adamgreig:
Which bootloaders? This seems like a bootloader bug, although we have the `set-sp` and `set-vtor` also just to work around bootloader bugs...

rhgndf:
I encountered this with the bootloader on py32f072 after flashing a firmware using its usb hid bootloader.

It seems like the bootloader enables the USB interrupt, SysTick and PendSV and doesn't disable it after jumping to user code, so if an interrupt fires before the user firmware data structures are set up, the interrupt handler can corrupt data or crash causing a stuck cpu.
The instructions in the bootloader before jumping are just:
```
  1fff01d2  ldr   r1,[r0,#4]       ; application reset handler
  1fff01d4  ldr   r0,[r0]          ; application initial MSP
  1fff01d6  msr   msp,r0
  1fff01da  bx    r1
```

The fix would be to enable both this feature and the set-vtor feature which seems to make it reliably boot.

jonathanpallant:
I think you could just disable interrupts, and write code in Rust inside main to reset your peripherals? I don't think restting all the interrupts needs to be pre-main, and I don't think it needs to be in cortex-m-rt, given it's only to work-around buggy bootloaders on specific platforms.

rhgndf:
There would still be a small window between setting vtor and disabling peripherals though. Putting it before setting vtor would be much less likely to crash since we assume that the data structures in ram are still set up for the bootloader's ISR.

Yeah I agree trying to fix every bootloader's bugs in cortex-m-rt isn't a good idea so I wanted to get some comments on this.

adamgreig:
I think it would probably be enough to have a new feature like `disable-interrupts` which just runs `cpsid i` at the start of reset, then the remaining code could happen in the user's main function? That keeps it to a short and easily understood feature and allows the most flexibility. It could even just be in a user-defined `pre_init` but I can see it being helpful to have as a feature in the same way we support set-sp, set-vtor, etc.

cortex-m-rt/src/lib.rs
```diff
@@ -556,6 +564,11 @@ cfg_global_asm! {
     ".cfi_startproc
      Reset:",
 
+    // If enabled, mask all configurable interrupts before any other startup code runs. This
+    // prevents interrupts left enabled by a bootloader from using the bootloader's vector table.
+    #[cfg(feature = "reset-interrupt-state")]
+    "cpsid i",
```
jannic, line 570: Would `cpsid f` be needed as well on targets that support it?
(Probably not - when interrupts are disabled, any HardFault between this point and the proper initialization of the vector table must be triggered by our own startup code, so it's our responsibility to not trigger such a fault in the first place. Unlike interrupts, that could be triggered by outside events.)

Dirbaio:
I think there's more state that this should reset:
- priorities (both for interrupts and exceptions)
- basepri
- faultmask

adamgreig:
> I think there's more state that this should reset:
> 
>     * priorities (both for interrupts and exceptions)
> 
>     * basepri
> 
>     * faultmask

What do you think of _only_ disabling interrupts (edit: I mean only `cpsid i`, all interrupts off, but not changing any interrupt control registers), then, so that user code can handle any required resets in main?

Dirbaio:
That would work too, I guess. but:
- It's less "batteries included", I like the idea of enabling one flag and having guaranteed clean nvic state.
- for cases where the bootloader is jumping to the app with irqs disabled (we just found out the stm32n6 does this over at Embassy) the state cleanup you want is precisely enabling irqs. If the flag is "reset nvic state" it does that for you, if it's just "disable irqs in early boot" it doesn't do much for you, you still need custom code to enable them.
- For setups like embassy-executor and RTIC that take over main, there might be no convenient place to do the cleanup.

PR #689, opened 2026-08-31
Mask the NVIC bitmap index so the accessors need no bounds check.

CordlessCoder:
`NVIC`'s interrupt-bitmap accessors index a 16-word array with `nr / 32`:
```rust
unsafe { (*Self::PTR).ispr[usize::from(nr / 32)].write(1 << (nr % 32)) }
```

`nr` comes from `InterruptNumber::number() -> u16`, so the optimiser thinks `nr / 32` is 0..2048 against an array of 16. Each accessor keeps a bounds check, a panicking branch, and a reference to `panic_bounds_check`.

Ten call sites across nine methods: `mask`, `unmask`, `is_enabled`, `is_active`, `is_pending`, `pend`, `unpend`, and the ARMv8-M `route_to_nonsecure`, `route_to_secure` and `is_routed_to_nonsecure`.
This replaces them with one private helper that masks the index to the array's own bound.

### Why this isn't a breaking change
No Cortex-M profile implements more than 496 external interrupts, so `nr / 32` is at most 15 on every architecture this crate supports and the mask never alters a legal index.

### Measurement

`embassy`'s `examples/rp` `multiprio` binary, rp2040, thumbv6m, release:
| variant | `.text` |
|---|---|
| control | 16340 B |
| this change | 16120 B |
| _delta_ | **−220 B** |

### DIFF
--- a/cortex-m/CHANGELOG.md
+++ b/cortex-m/CHANGELOG.md
@@ -9,12 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - `NVIC`'s interrupt-bitmap accessors no longer emit a bounds check and a panic
-  path. `mask`, `unmask`, `is_enabled`, `is_active`, `is_pending`, `pend`,
-  `unpend` and the ARMv8-M `route_to_*`/`is_routed_to_nonsecure` all indexed a
-  16-word array with `nr / 32`, which the optimiser cannot prove is in range for
-  a `u16`. No Cortex-M implements more than 496 external interrupts, so the
-  index is now masked to the array's own bound, which no reachable interrupt
-  number reaches.
+  path.
 
 ## [v0.7.9] - 2026-08-18
 
--- a/cortex-m/CHANGELOG.md
+++ b/cortex-m/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- `NVIC`'s interrupt-bitmap accessors no longer emit a bounds check and a panic
+  path. `mask`, `unmask`, `is_enabled`, `is_active`, `is_pending`, `pend`,
+  `unpend` and the ARMv8-M `route_to_*`/`is_routed_to_nonsecure` all indexed a
+  16-word array with `nr / 32`, which the optimiser cannot prove is in range for
+  a `u16`. No Cortex-M implements more than 496 external interrupts, so the
+  index is now masked to the array's own bound, which no reachable interrupt
+  number reaches.
+
 ## [v0.7.9] - 2026-08-18
 
 ### Fixed
--- a/cortex-m/src/peripheral/nvic.rs
+++ b/cortex-m/src/peripheral/nvic.rs
@@ -321,7 +321,7 @@ impl NVIC {
     /// The `&` helps remove an expensive bounds check - we know all valid interrupt numbers will be in range.
     #[inline]
     const fn reg_index_and_mask(irq: u16) -> (usize, u32) {
-        debug_assert!(irq < 512);
+        debug_assert!((irq as usize) < (Self::BITMAP_WORDS * 32));
         let idx = (irq / 32) as usize & (Self::BITMAP_WORDS - 1);
         let mask = 1 << (irq % 32);
         (idx, mask)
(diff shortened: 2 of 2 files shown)


cortex-m/CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
```
jonathanpallant, line 10: I think this CHANGELOG entry can be reduced to the first sentence.

cortex-m/src/peripheral/nvic.rs
```diff
@@ -82,6 +82,14 @@ pub struct RegisterBlock {
     pub stir: WO<u32>,
 }
 
+/// `BITMAP_WORDS` describes the arrays above rather than being asserted about itself: the gap from
+/// `ISER` to `ICER` is one bitmap plus one reserved bitmap of the same size, so resizing either
+/// without the other fails this.
+const _: () = assert!(
```
jonathanpallant: This doesn't seem useful? Why not just define the fields as being `pub iser: [u32; IRQ_BITMAP_WORDS]` etc?

cortex-m/src/peripheral/nvic.rs
```diff
@@ -305,12 +313,28 @@ impl NVIC {
...
+    /// and `ITNS`.
+    const BITMAP_WORDS: usize = 16;
+
+    /// Which 32-bit word of an interrupt bitmap holds `nr`'s bit.
+    ///
+    /// The mask is what lets this be indexed without a bounds check. Each bitmap is
+    /// [`BITMAP_WORDS`](Self::BITMAP_WORDS) words, spanning 512 interrupts; no Cortex-M implements
+    /// more than 496, so `nr / 32` is at most 15 and the mask never alters a legal index. Without
+    /// it the compiler only knows `nr` is a `u16`, so `nr / 32` could be up to 2047, and every
+    /// accessor below keeps a bounds check and a panic path that no reachable input can take.
```
jonathanpallant: The accessors are above. But also, this could be reduced to:

```rust
/// Which 32-bit word of an interrupt bitmap holds `nr`'s bit.
///
/// The mask helps remove an expensive bounds check - we know all valid interrupt numbers will be in range.
```

cortex-m/src/peripheral/nvic.rs
```diff
@@ -305,12 +313,28 @@ impl NVIC {
...
+    #[inline]
+    const fn reg_index(nr: u16) -> usize {
+        (nr / 32) as usize & (Self::BITMAP_WORDS - 1)
```
jonathanpallant: This function could also return the mask (the `1 << (irq % 32)`) to save that code being repeated. It still gets optimised out nicely: https://rust.godbolt.org/z/7Kn8aYb66

cortex-m/src/peripheral/nvic.rs
```diff
@@ -305,12 +307,26 @@ impl NVIC {
...
         unsafe { ((*Self::PTR).itns[group_idx].read() & bit_mask) == bit_mask }
     }
 
+    /// Words in each of the NVIC's interrupt bitmaps: `ISER`, `ICER`, `ISPR`, `ICPR`, `IABR`
+    /// and `ITNS`.
+    const BITMAP_WORDS: usize = 16;
+
+    /// Which 32-bit word of an interrupt bitmap holds `nr`'s bit.
+    ///
+    /// The mask helps remove an expensive bounds check - we know all valid interrupt numbers will be in range.
```
jonathanpallant: Sorry, I used the work *mask* twice in my sample with two different meanings.

Perhaps:

```suggestion
    /// The `&` helps remove an expensive bounds check - we know all valid interrupt numbers will be in range.
```

jonathanpallant: You could also replace `512` with `BITMAP_WORDS * 32` in the assert.

diondokter (approved):
Looks good. Nice and straightforward now

adamgreig (approved):

adamgreig:
Just to confirm, @CordlessCoder you still see the same binary size improvement and lack of panic path in the current version of the PR?

CordlessCoder:
@adamgreig Yes, `panic_bounds_check` is still absent from the version with the PR and present without it. Same -220B flash difference.

PR #690, opened 2026-09-04
Fixes #642 Remove redundant non-cortex-m stub for interrupt::free

### LINKED ISSUE #642: Cleanup free functions
My asm refactor allows us to get rid of https://github.com/rust-embedded/cortex-m/blob/6baaf6ab1c0a8cf22ae2e0b895b077e81eddecfd/cortex-m/src/interrupt.rs#L90-L102 and remove the [cfg guard from the normal free function](https://github.com/rust-embedded/cortex-m/blob/6baaf6ab1c0a8cf22ae2e0b895b077e81eddecfd/cortex-m/src/interrupt.rs#L67).

The normal free function is already compileable on non-cortex-m because it relies on functions that will simply call `unimplemented!()` if they are built on unsupported targets.

jotheesh1729:


cortex-m/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Removed
+- The separate `#[cfg(not(cortex_m))]` stub for `interrupt::free`. The real
```
jonathanpallant: I don't think we need a lengthy justification in the CHANGELOG. The first sentence is fine - the PR and the commit history record the rest.

jotheesh1729: done

jonathanpallant (approved):
I don't think this is of high importance, but I think it's fine.

